Issue #3, opened 2016-02-05
colors not showing up


tnguyen14:
I added the colors for `diff-highlight` in my `.gitconfig` as instructed, however the output of `git diff | diff-highlight` does not output any colors. 

Could there be something that I am missing here?

stevemao:
`git diff --color | diff-highlight` should work.
But I use `git config --global core.pager "diff-highlight | diff-so-fancy | less -r"` so any diff related commands will use `diff-so-fancy` :)

tnguyen14:
Thank you. Using the `--color` flag worked.

When I tried the config to setting `core.pager` as you suggested, I see an issue with indentation being off, as such:

```diff
@@ -68,6 +68,7 @@
        whitespace = space-before-tab,indent-with-non-tab,trailing-space
        quotepath = false
        editor = vim
  pager = diff-highlight | diff-so-fancy | less -r
        # editor = subl -n -w
```

stevemao:
Can you open another issue if it's not related to your original one? It would be helpful if you could include a screenshot. Thanks.

PR #8, opened 2016-02-08
readme: add --color flag to `git diff` command

tnguyen14:
address issue in #3


stevemao:
This is just an example though.

tnguyen14:
Do you mean that the `--color` flag is not mandatory?

stevemao:
all flags are optional. But to work with `diff-so-fancy` we actually need colour. Thanks.

PR #11, opened 2016-02-09
colorize empty lines

noscript:
Due to hidden `+` and `-` it's impossible to see removed and added empty lines. This PR addresses this issue.

Before:
before

After:

after

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -5,6 +5,7 @@
 color_code_regex=$'(\x1B\\[([0-9]{1,2}(;[0-9]{1,2})?)[m|K])?'
 reset_color="\x1B\[m"
 dim_magenta="\x1B\[38;05;146m"
+invert_color="\x1B\[7m"
 
 format_diff_header () {
 	# simplify the unified patch diff header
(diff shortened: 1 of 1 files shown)


paulirish:
Oh that's a really nice addition. :) thanks!

noscript:
Requires some more testing and will be merged eventually.

stevemao:
I had the same problem. Thanks @noscript :)

finishingmove:
Hey guys, this would be a welcome feature. How about painting just one brick (column) instead of the whole row though?

stevemao:
I don't know how hidden unicode characters work here with the colour.

noscript:
Coloring just one brick is possible. I can probably add a flag to color whole like or one column.

> I don't know how hidden unicode characters work here with the colour.

There is no hidden unicode, at least not used for empty line coloring. It's a normal space character.

stevemao:
@noscript Sorry ignore my silly question. It's only for adding or deleting empty lines.

finishingmove:
The concern about my suggestion is how this representation would be differentiated from a line where a single white space was added or removed. So, maybe it wasn't such a great idea after all? ;)

stevemao:
@finishingmove Good point :), my original concern is about whitespaces (or hidden unicode characters).

stevemao:
@noscript I did some testing, the best way of doing it is probably use the brick (first whitespace) of each line. It was reserved for `+` and `-` that's removed by this script :) I'm testing on Windows and let me know if it's not the case on Mac.

stevemao:
Hi @noscript why did you close this?

noscript:
I see no reason to have two competing PRs.

stevemao:
They are not competing. We just need to discuss and see what's the best way to do this.

paulirish:
In general, +1 to having competing PRs. diff-so-fancy is opinionated, so it's fine for us to discuss alternatives

<hr>

Now for this PR, I tried it out and found some issues (similar to recent issues found in #23):

image

However even if fixed, I think the highlight across the full line is too strong, and I prefer the single block highlight instead.

PR #13, opened 2016-02-09
depend on local diff-highlight.

### LINKED ISSUE #6: Hard dependency on diff-highlight from diff-so-fancy

paulirish:
fixes #6 

hows this look to everyone?


paulirish:
@noscript what do you think?

aziz:
:+1: Hiding the complexity from end user.

stevemao:
This looks great :+1:

noscript:
I think it's better to check first if there is already `diff-highlight` in `$PATH` and if not then fallback to the one in `./third_party`.

stevemao:
@noscript since we are adding `update-deps.sh` its more likely that our `diff-highlight` is probably more updated.
Also if we only hard depend on our own `diff-highlight` and we have more control. Say there might be a mismatch of `diff-highlight` on user's machine and we have to ask user manually fix their dependency.

noscript:
Ignore my previous comment, that won't work. Version of `diff-highlight` that comes with `git` package has no executable flag.

> @noscript since we are adding update-deps.sh its more likely that our diff-highlight is probably more updated.

A side issue is that before this PR `diff-so-fancy` didn't have an internal dependency on `diff-highlight`. One could simply drop `diff-so-fancy` anywhere in $PATH and configure git pager.

After this PR it becomes less convenient for non-npm users, since `diff-so-fancy` will be looking for hardcoded `./third_party` directory.

This PR can be merged as is now and later I can introduce a flag for `diff-so-fancy` to fallback to old behavior.

update-deps.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+curl -o "third_party/diff-highlight/diff-highlight" "https://raw.githubusercontent.com/git/git/master/contrib/diff-highlight/diff-highlight"
+curl -o "third_party/diff-highlight/README" "https://raw.githubusercontent.com/git/git/master/contrib/diff-highlight/README"
```
stevemao: Can we just copy the whole folder or use glob instead of `curl`ing twice?

noscript: Not sure about `curl` but with `wget` we can do
```
wget -N -P third_party/diff-highlight https://raw.githubusercontent.com/git/git/master/contrib/diff-highlight/{diff-highlight,README}
```

stevemao:
> A side issue is that before this PR diff-so-fancy didn't have an internal dependency on diff-highlight.

This is not true. If the npm bin takes higher precedence over user's own path that contains `diff-highlight`, user will always run our `diff-highlight`. if the other way, user will never run our own `diff-highlight`.



> later I can introduce a flag for diff-so-fancy to fallback

Like I said there might be a mismatch. `npm` uses local dependencies over global to avoid that. `gem` uses the global approach and I heard a lot of people having trouble. If we do introduce such a flag it'll be only for advanced users and we won't support that (Say user has v1 and our script depends on v2, its too much overhead to make it work on both for too little gain).

And, like @paulirish already mentioned, diff-highlight doesn't update very often so I wonder it's worth the effort...

update-deps.sh
paulirish: wget isnt on mac by default, so i hesitate to use that.

cht8687:
This PR not only changes the repo to a shell project but also clarifies the third-party dependency, quite good move!!! :+1:

PR #23, opened 2016-02-10
mark empty newlines

stevemao:
color the brick (first white space) of each line that is reserved for "+" and "-" that's removed by this script
Alternative of and inspired by #11

diff-so-fancy

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -13,6 +13,7 @@ fi
 color_code_regex=$'(\x1B\\[([0-9]{1,2}(;[0-9]{1,2})?)[m|K])?'
 reset_color="\x1B\[m"
 dim_magenta="\x1B\[38;05;146m"
+invert_color="\x1B\[7m"
 
 format_diff_header () {
 	# simplify the unified patch diff header
@@ -28,9 +29,13 @@ colorize_context_line () {
 	$SED -E "s/@@$reset_color $reset_color(.*$)/@@ $dim_magenta\1/g"
 }
 
+mark_empty_lines () {
+	$SED -E "s/^$color_code_regex[\+\-]$reset_color\s*$/$invert_color\1 $reset_color/g"
+}
+
 strip_leading_symbols () {
 	# strip the + and -
-	$SED -E "s/^($color_code_regex)[\+\-]/\1 /g"
+	$SED -E "s/^$color_code_regex[\+\-]/\1 /g"
 }
 
 print_horizontal_rule () {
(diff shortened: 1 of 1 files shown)


paulirish:
tried out this patch. I'm seeing the inversion applying to text in some cases:

image

stevemao:
@paulirish just pushed the fix. On my Mac I need to close the ansi but not on my Windows...

stevemao:
Oh no there is still some problem :(

paulirish:
Yup, I just saw those. :)   

Those issues fixed, though I found this: 

image

noscript:
You reset the color too early. That should  fix it:
```diff
-	sed -E "s/^($color_code_regex)[\+\-]($reset_color)\s*$/$invert_color\2 \1/g"
+	sed -E "s/^($color_code_regex)[\+\-]($reset_color)\s*$/$invert_color\1 \2/g"
```

stevemao:
I have a feeling, we might be using different versions of `sed`. I updated it on my mac and it works fine (with all the patch suggestions here).

My sed version is 4.2.2

noscript:
Mine is sed (GNU sed) 4.2.2

stevemao:
The `\2` or `\1` here gives me the same results. So I could just do:

```
sed -E "s/^($color_code_regex)[\+\-]$reset_color\s*$/$invert_color\1 \1/g"
```

stevemao:
I think we should always test against the latest sed (GNU sed)

Can you guys verify these patches work for you (all the commits here should work)? I don't really understand why they all work I just blindly tested. It's hard to test these invisible ansi escapes. Can't wait until we have tests (#32)

diff-so-fancy
```diff
@@ -22,9 +23,13 @@ colorize_context_line () {
...
 }
 
+mark_empty_lines () {
+	sed -E "s/^($color_code_regex)[\+\-]$reset_color\s*$/$invert_color\1 /g"
+}
+
 strip_leading_symbols () {
 	# strip the + and -
-	sed -E "s/^($color_code_regex)[\+\-]/\1 /g"
+	sed -E "s/^($color_code_regex)[\+\-]/\1/g"
```
stevemao: This change is not needed

noscript:
> The `\2` or` \1` here interchangeable for me the same results.

Even giving the same result, these are not interchangeable really.

stevemao:
The reason is the ansi escape character is captured twice...

stevemao:
@paulirish I understand now. 
It's because your using ` git config --global core.pager "diff-highlight | diff-so-fancy | less -r"`
if you use `git config --global core.pager "diff-so-fancy | less --tabs=1,5 -R"` (the one on readme) it looks fine.

stevemao:
Hope my last commit fixes it.

elyscape:
The screenshot in the readme should probably be updated to reflect this.

stevemao:
@elyscape Just send a PR

elyscape:
@stevemao I can't figure out how the screenshot was hosted, otherwise I would.

PR #24, opened 2016-02-10
Fix diff-highlight path detection for npm install

clkao:


stevemao:
Why? Note that the current master is not published to npm yet.

clkao:
Oh nvm, was modifying the script in the install path and got confused.

anshul:
This is probably needed on systems that use osx nvm.

PR #25, opened 2016-02-10
Detect gsed and use it if it exists

anshul:


anshul:
@stevemao Squashed the changes into a single commit.

stevemao:
Is it common to symlink `gsed` to `sed`? homebrew does it. Maybe we should instead warn the user?

anshul:
This is a common pattern on osx/homebrew.   `brew install sed find readlink` will create `gsed`, `gfind`, `greadlink`.  Overriding the apple implementations of these commands can give unpredictable behaviour for various apple software.  I never use or recommend `--with-default-names`

anshul:
http://apple.stackexchange.com/questions/69223/how-to-replace-mac-os-x-utilities-with-gnu-core-utilities

anshul:
Sorry, I read your comment wrong.  Edited mine.  I think, everyone is served better if npm install just works out of the box on a typical osx install.  Recommending overriding the apple sed command could leave other commands that depend on the apple installation in a bad shape.

anshul:
Looks like ports also installs a gsed. http://apple.stackexchange.com/a/166593

stevemao:
I'm fine with this. I'll see what other people think.

paulirish:
Yup. This is good. Osx sed is crap but some people won't want to replace it and that's fine.

diff-so-fancy
```diff
@@ -3,6 +3,10 @@
 [ $# -ge 1 -a -f "$1" ] && input="$1" || input="-"
 
 diff_highlight="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )/third_party/diff-highlight/diff-highlight"
+SED=sed
+if hash gsed 2>/dev/null; then
+  SED=gsed
+fi
```
paulirish: Can you make this a one liner?

anshul: Sure

anshul:
@paulirish Does that one liner look fine?

paulirish:
Yup! Ace.

PR #29, opened 2016-02-10
use which to find our dependency

### LINKED ISSUE #6: Hard dependency on diff-highlight from diff-so-fancy

stevemao:
Note our `diff-highlight` is not the hard dependency. It's there for convience so the script is in $PATH. It's not possible to guarantee to know where exactly our `diff-highlight` is.

Fixes #6

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -2,7 +2,7 @@
 
 [ $# -ge 1 -a -f "$1" ] && input="$1" || input="-"
 
-diff_highlight="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )/third_party/diff-highlight/diff-highlight"
+diff_highlight=$(which diff-highlight)
 
 color_code_regex=$'(\x1B\\[([0-9]{1,2}(;[0-9]{1,2})?)[m|K])?'
 reset_color="\x1B\[m"


paulirish:
Not sure about this change. It will only identify a global diff-highlight which our manual installer friends won't have.  I'd rather directly depend on our local dep.

stevemao:
How to identify where our local `diff-highlight` is?

noscript:
That's bad solution:
```shell
diff_highlight=$(which diff-highlight)
```

Probably you don't want to have `diff_highlight` to be assigned to `diff-highlight not found`. One need to check the exit code first.

paulirish:
> How to identify where our local diff-highlight is?

What's wrong with what you had?

```sh
diff_highlight="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )/third_party/diff-highlight/diff-highlight"
```

I'd think we could augment that by using a global `diff-highlight` if that file isn't found. But otherwise that seems ideal

stevemao:
when you do `npm install` or `npm link`, it puts both files in a global `bin`, so `diff-highlight` is not in `/third_party/diff-highlight` anymore.

paulirish:
@stevemao good call.  (btw thx for fixing my bug with the BASH_SOURCE fix earlier)

Current manual install instructions say put both scripts in your path, but that's kinda out of date. Ideal manual install, IMO, is user clones the repo and adds a symlink to `diff-so-fancy` to somewhere in PATH. Does that seem best? 

If so, then we should just use the combo of `which` followed up by local fallback.

stevemao:
Yes but as @noscript mentioned earlier, 
> Version of diff-highlight that comes with git package has no executable flag.
 you don't want to have diff_highlight to be assigned to diff-highlight not found. 

We need to be mindful for those points.

paulirish:
> Version of diff-highlight that comes with git package has no executable flag.

yeah we would never use this. if we find a global with `which` then it was installed with npm. it has executable flag. if someone symlinked their own, then its their responsibility to chmod. 

> you don't want to have diff_highlight to be assigned to diff-highlight not found.

it wouldn't be. we'd check the error code and use our fallback.

paulirish:
I'm following up with a PR…

stevemao:
Closing in favor of #39

Issue #22, opened 2016-02-10
Have a way to switch between old way of `git diff` and the new one


prat0318:
Sometimes, i do want to get a patch from `git diff`. Is there an easy way to do that once i have turned on `diff-so-fancy` git config?

stevemao:
```
$ git config --global core.pager less
```

paulirish:
There's probably a less destructive way. I'm curious as well. Let's document a decent solution in the readme

paulirish:
Easy enough!  `git --no-pager diff`  :)

PR #30, opened 2016-02-10
Add git alias usage to docs

tommarshall:
Thanks for `diff-so-fancy`. It's a great idea :+1:

This change adds a git alias definition for `diff-so-fancy` to the 'Usage' section of the README.

A git alias has the benefit of requiring less keystrokes than the piped series of commands each time, is tab-completed by git's default autocompletion, and leaves the default behaviour of `git diff` intact.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -24,6 +24,11 @@ git diff --color | diff-so-fancy
 git config --global core.pager "diff-so-fancy | less --tabs=1,5 -R"
 ```
 
+Or, create a git alias  in your `~/.gitconfig` for shorthand fanciness:
+```shell
+diff-so-fancy = "!git diff --color $@ | diff-so-fancy"
+```
+
 ## Install
 
 For convenience, the recommended installation is via NPM:


noscript:
Not applicable since #13.  Now `diff-highlight` is called from withing `diff-so-fancy`.

tommarshall:
Aah thanks. I'd missed that.

I think the following alias could still be helpful to people.

```sh
diff-so-fancy = "!git diff --color $@ | diff-so-fancy"
```

The key strokes are similar but the autocomplete makes it easier to reach for than the piped approach. And an alias is a useful approach for users that don't want to override the default `diff` behaviour, e.g. #22.

tommarshall:
I've applied the change. Thanks for spotting @noscript.

paulirish:
Seems good.  Thanks!

readme.md
```diff
@@ -24,6 +24,11 @@ git diff --color | diff-so-fancy
 git config --global core.pager "diff-so-fancy | less --tabs=1,5 -R"
 ```
 
+Or, create a git alias  in your `~/.gitconfig` for shorthand fanciness:
+```shell
+diff-so-fancy = "!git diff --color $@ | diff-so-fancy"
```
stevemao: the alias `diff-so-fancy` is the same as our `diff-so-fancy`, not sure if this would confuse people.

tommarshall: That's fair. Happy to change it to whatever you feel is best. Would you prefer `git dsf` for the alias?

stevemao: @paulirish @noscript what do you think?

noscript:
Purpose of alias is to type less, for that `diff-so-fancy` is too long. In that case `dfs` is better.

readme.md
noscript: Purpose of alias is to type less, for that diff-so-fancy is too long. In that case dfs is better.

PR #31, opened 2016-02-10
remove src/dest prefix for cmd-click support. fixes #28

### LINKED ISSUE #28: Unable to quickly select a filename
When double clicking the filename the `a/` or `b/` is copied along. It would be great to remove that, or separate it with a space, or add the plain filename as a separate line as well, to be able to support this use case.

paulirish:
screenshots: 
image

image

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -13,8 +13,9 @@ format_diff_header () {
 	sed -E "s/^($color_code_regex)diff --git .*$//g" | \
 	sed -E "s/^($color_code_regex)index .*$/\
 \1$(print_horizontal_rule)/g" | \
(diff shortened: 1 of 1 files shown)


noscript:
regex are expensive and in my opinion being overused, specially for something that can carried out by git.

just my 2c

stevemao:
I do notice it makes my PC slower... But doesn't affect my new Mac :)

paulirish:
I definitely prefer having links be cmd-clickable. 
And while users can add `--no-prefix`, we can't provide that as a feature.

Let's keep #28 open but I'll close this PR. I'd like to find an alternate and way of handling this modification.

PR #33, opened 2016-02-10
Define shorter `dsf` alias in docs

tommarshall:
Shorten the docs git alias to `dsf` as per the discussion in: https://github.com/stevemao/diff-so-fancy/pull/30


paulirish:
I was going to suggest we document the command instead like `git config --global alias.dsf "!git diff --color $@ | diff-so-fancy"`  but on second thought, the config line is more reasonable.

This shorthand works for me. Thumbs up, guys?

stevemao:
:+1:

PR #36, opened 2016-02-10
Fixes path detection for nvm users

anshul:
nvm creates the symlinks:
  * ~/.nvm/versions/node/v5.5.0/bin/diff-so-fancy -> ~/.nvm/versions/node/v5.5.0/lib/node_modules/diff-so-fancy/diff-so-fancy
  * ~/.nvm/versions/node/v5.5.0/bin/diff-highlight -> ~/.nvm/versions/node/v5.5.0/lib/node_modules/diff-so-fancy/third_party/diff-highlight/diff-highlight

which was breaking the path detection.

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -2,7 +2,7 @@
 
 [ $# -ge 1 -a -f "$1" ] && input="$1" || input="-"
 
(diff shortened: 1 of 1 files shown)


anshul:
This fixes the same problem as https://github.com/stevemao/diff-so-fancy/pull/24

stevemao:
What do you think of #29?

paulirish:
This works for me. Let's merge and then followup with whats happening in #29

PR #37, opened 2016-02-10
Fix color code regex to match all bash color codes

xixixao:
This was breaking default mercurial output which is `0;32;1`. Theoretically you could have more than three codes in the sequence, but I haven't seen that used.


stevemao:
Stupid question: I don't know if this is valid ansi-escape. 

If so, `(;[0-9]{1,2})?` part could be changed to something like `(;[0-9]{1,2}){0,1,2}`?

stevemao:
Looks correct by my eyes balls. Can you rebase?

paulirish:
i rebased and merged.

xixixao:
Thanks @paulirish

PR #38, opened 2016-02-10
Don't require index line to render correctly

xixixao:
Mercurial doesn't output the `index a7f8d9e..3fa2821 100644` line, so this PR removes the line completely (that sed wouldn't work on two index lines after each other, but that's not gonna happen in the diff output) and prepends the horizontal rule tu the `---` file line instead.

It's a bit cryptic, so if you have a solution for mercurial, let me know, or if you don't care enough, don't merge it.

Needs careful merge with the other PR I put up (since the back reference numbers change in that one).

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -8,11 +8,14 @@ color_code_regex=$'(\x1B\\[([0-9]{1,2}(;[0-9]{1,2})?)[m|K])?'
 reset_color="\x1B\[m"
 dim_magenta="\x1B\[38;05;146m"
 
+git_index_line_pattern="^($color_code_regex)index .*"
+
 format_diff_header () {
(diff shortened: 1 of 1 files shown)


paulirish:
I like the thrust of this. Let's merge both PRs. 

Can you rebase them?  (sorry!)

finishingmove:
I noticed that the first line of `git log` output is not visible until I scroll a bit down, then back to the top again. Is this related or should I open a new issue?

xixixao:
@finishingmove With this PR merged in?

finishingmove:
@xixixao No, before. That's why I'm asking. Did anyone else experience that issue?

paulirish:
merged

PR #39, opened 2016-02-10
prefer a global diff-highlight, but fallback to our local copy

### LINKED ISSUE #6: Hard dependency on diff-highlight from diff-so-fancy

paulirish:
This builds on the discussion in #29 and fixes #6.
All the context is there.

It also fixes a small bug in @anshul's pr #36 . :)

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -2,9 +2,13 @@
 
 [ $# -ge 1 -a -f "$1" ] && input="$1" || input="-"
 
-diff_highlight="$( cd "$( dirname realpath "${BASH_SOURCE[0]}" )" && pwd )/third_party/diff-highlight/diff-highlight"
-
 hash gsed 2>/dev/null && SED=gsed || SED=sed
+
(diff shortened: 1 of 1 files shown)


stevemao:
Should it prefer global? I was thinking:

1. find our local hard dependency, if not found
2. find global dependency (may not be our hard dependency), if not found
3, throw error

Dependency needs to be executable.

bquorning:
:+1: Thank you. #36 broke diff-so-fancy for me, and this PR fixes it again.

paulirish:
Cool. Steve, I'm fine with s follow-up commit to swap the order

anshul:
Oops.  @paulirish Thanks for fixing.

stevemao:
Oh well only you think it's better @paulirish :)

paulirish:
I don't have a strong feeling on the order

stevemao:
Cool. Don't worry about it then.

Issue #35, opened 2016-02-10
Git add --patch does not use diff-so-fancy


mrageh:
Hi 

Not sure if this is the correct place for this issue.

Just started using this npm package and it's great, it makes reading diffs in my terminal great. 

However when I use `git add --patch` it uses the old style diff to display each chunk that's about to be staged. 

Is `git add --patch` not supposed to use `diff` under the hood, which is now `diff-so-fancy`?

Is there anything we can do to use `diff-so-fancy` with `git add --patch`?

Thanks

paulirish:
Someone else pinged me about this a month ago. I'm not sure. 

Based on https://github.com/git/git/blob/master/git-add--interactive.perl#L747-L768 it doesn't look good.

But i'd love to be proven wrong. There's a lot of mention of color in here, so there may be SOME configurability, but not sure if we can intercept with a pager or something.

paulirish:
There's some other people who were poking around this:

http://www.nickcoding.com/2014/10/30/making-git-prose-friendly/
http://git.661346.n2.nabble.com/Different-diff-strategies-in-add-interactive-td7588815.html

.. which my unveil some solutions.

mrageh:
Thanks for the swift response @paulirish, would it be worth leaving this issue open or do you think we should close it?

paulirish:
Leave it open. It still requires some research to see if it's possible or not.

tonekk:
:+1: was just going to add the same issue

danielcompton:
`git commit --verbose` doesn't use it either.

hultberg:
As @danielcompton pointed out, `git commit -v` does not use diff-so-fancy.

paulirish:
The solution here appears to be to fork `$brewdir/Cellar/git/2.7.0/libexec/git-core/git-add--interactive` and make some edits.

I don't see a cleaner way to configure this.

hultberg:
We could open a feature request to git itself and see if we could get a config for this?

paulirish:
> We could open a feature request to git itself and see if we could get a config for this?

Yup, we'll need to get this supported upstream in git core first. 

<hr>

@peff I think you're the right person to help us out. First of all, thank you. This project has been a big success mostly due to your work on diff-highlight. 

[...]

PR #44, opened 2016-02-12
Fixes #42: Don't use diff-so-fancy for core pager

### LINKED ISSUE #42: Current git config instructions clobbers `git log`
`git config --global core.pager "diff-so-fancy | less --tabs=1,5 -R"`

This highjacks things like `git log`, and other commands locally for me. Is there a non-destructive way of doing this? If not, I'd suggest showing how to alias it in bash instead of adding it to the gitconfig.

noscript:
Addresses #42.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -19,9 +19,10 @@ You can do one-off fanciness:
 git diff --color | diff-so-fancy
 ```
 
-**But**, you'll probably want to fancify all your diffs. Run this so `git diff` will use it:
+**But**, you'll probably want to fancify all your diffs. Run this so `git diff` and `git show` will use it:
 ```shell
-git config --global core.pager "diff-so-fancy | less --tabs=1,5 -R"
+git config --global pager.diff "diff-so-fancy | less --tabs=1,5 -R"
+git config --global pager.show "diff-so-fancy | less --tabs=1,5 -R"
 ```
 
 Or, create a git alias  in your `~/.gitconfig` for shorthand fanciness:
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -21,7 +21,8 @@ git diff --color | diff-so-fancy
 
 **But**, you'll probably want to fancify all your diffs. Run this so `git diff` will use it:
 ```shell
-git config --global core.pager "diff-so-fancy | less --tabs=1,5 -R"
+git config --global pager.diff "diff-so-fancy | less --tabs=1,5 -R"
```
stevemao: I'd say something like "If you only want to use it with git-diff, run blabla".

stevemao: Oh I see, this is trying to avoid git-log using the same pager.

stevemao:
LGTM, is there any other command that might work great with our script?

noscript:
`git add --patch` produces diff output but I'm not sure how to set a pager for that.

readme.md
```diff
@@ -71,7 +72,7 @@ You may also want to configure [general diff colors](https://github.com/pauliris
 Sometimes you will want to bypass diff-so-fancy. Easy enough:
 
 ```shell
-git --no-pager diff  # will avoid core.pager hook
+git --no-pager diff  # will avoid pager.diff hook
```
stevemao: I think this should say "will avoid any pager hook"

readme.md
```diff
@@ -68,10 +69,10 @@ You may also want to configure [general diff colors](https://github.com/pauliris
 
 ## Opting-out
 
-Sometimes you will want to bypass diff-so-fancy. Easy enough:
+Sometimes you will want to bypass diff-so-fancy. Use `--no-pager` for that:
```
stevemao: I prefer the original wording but I'll let @paulirish to make the call :)

paulirish:
Yup. This is the best way to handle it.

vassudanagunta:
> `git add --patch` produces diff output but I'm not sure how to set a pager for that. #35

Not possible I think. See my answer to [How do I use a pager for long git add --patch hunks?](https://stackoverflow.com/questions/39743490/how-do-i-use-a-pager-for-long-git-add-patch-hunks/63652969#63652969). Does anyone disagree?

PR #50, opened 2016-02-12
Add `-F -X` flags to the pager so that less doesn't clear the screen for a short diff

arthurdarcet:
`-F` makes less return immediately if the content fits on the screen. Without this you need to press Q to get back to your shell even if the diff was only a few lines long
`-X` prevent less from clearing the screen at the beginning. Which is undesirable if the content fits on one screen and unnecessary if the content is longer than a screen.

And the `-R` flag (not added by this PR) displays colors correctly instead of outputting as ASCII representation of the control chars

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -21,8 +21,8 @@ git diff --color | diff-so-fancy
 
 **But**, you'll probably want to fancify all your diffs. Run this so `git diff` and `git show` will use it:
 ```shell
-git config --global pager.diff "diff-so-fancy | less --tabs=1,5 -R"
-git config --global pager.show "diff-so-fancy | less --tabs=1,5 -R"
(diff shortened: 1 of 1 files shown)


vanniktech:
:+1: I added this one locally

paulirish:
`less --tabs=1,5 -R -F -X` is kinda insightful even if it's the best thing to do. 

Just curious, can we do `less -RFX --tabs=1,5`

Also, can you just state here what the RF and X flags do, for people that end up landing here?

paulirish:
Also, I just merged in a conflict for you. Might want to rebase.

stevemao:
:+1:

arthurdarcet:
Rebased on master and documented the flags (here and in the commit message)

paulirish:
you're a pro. thanks @arthurdarcet

PR #51, opened 2016-02-12
Strip spaces

xixixao:
The point of stripping the symbols is to be able to easily copy and paste code. That is not currently ideal because there is a single space at the beginning of each line left.

Also, for the love of me, don't give an unsuspecting stranger commit rights :D

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -40,7 +40,7 @@ mark_empty_lines () {
 
 strip_leading_symbols () {
 	# strip the + and -
-	$SED -E "s/^$color_code_regex[\+\-]/\1 /g"
+	$SED -E "s/^$color_code_regex[\ \+\-]/\1/g"
 }
 
 print_horizontal_rule () {


paulirish:
This is true. The indentation is nice for readability, but it defeats the copy/pastable-ness of the output. 

Is there any way to inject indentation that would not be copied to the clipboard? (Some ANSI escape perhaps?)

xixixao:
Not that I am aware of. Why do you think a single space is nice for readability? I have everything that's not diffed in colors (modifed file, the files header, the line numbers). On the other hand, this doesn't have to be part of the tool, it's an easy `sed`. But I thought it might be useful for everyone.

stevemao:
For adding/removing a newline, a marked single space can make it different. If a single space is stripped and I assume we keep the mark, its impossible to tell if I added/removed a newline or if a added//removed a single space at the beginning of an existing line. But that's an edge case.

xixixao:
I haven't tested this with markers for empty lines. They should be inserted after the initial space.

stevemao:
@xixixao Oh that's not a bad idea :)

xixixao:
But this won't be ideal as those spaces will show up in the copied output - goes back to Paul's suggestion of uncopyable character. Still probably better than what we have now.

xixixao:
Probably the best we can do - it doesn't bother me much, since my editors strip trailing whitespace on save.

stevemao:
From my testing, all unicode characters can be copied/pasted. some are hidden so it's even worse. Maybe we should use a different colour to mark newlines (say dim).

stevemao:
Closing in favor of #62

PR #52, opened 2016-02-12
simplify realpath/bash_source logic

paulirish:
Realpath isn't present on a lot of systems and I believe would necessitate extra dependencies.

The included technique is courtesy of https://gist.github.com/TheMengzor/968e5ea87e99d9c41782 and works well in my tests.


paulirish:
cc @anshul . please take a look

stevemao:
This is the same as the first answer of http://stackoverflow.com/questions/59895/can-a-bash-script-tell-what-directory-its-stored-in

PR #53, opened 2016-02-12
Add test suite

### LINKED ISSUE #32: Test suite
I think it's time for us to add proper unit tests.

I haven't seen a test suite for a project like this before, but it shouldnt be too bad. I'm thinking a folder of input text files, with ANSI escapes and all. And a folder of expected output. Seem right?

Is there a test runner that'd work well for us?

paulirish:
* Test suite setup with [Bats](https://github.com/sstephenson/bats)
* Assertion library courtesy of the very well done [bats-assert](https://github.com/ztombol/bats-assert)

## Fixtures
I've set up a few fixtures, based on some tricky cases. 
```sh
# creating a fixture file:
git --no-pager diff .  > fixture.diff
# running it directly:
cat fixture.diff | ./diff-so-fancy
```

## Assertions

The basics look like this:

```sh
output=$( load_fixture "ls-function" | $diff_so_fancy )

@test "that it doesnt break" {
  assert_output --partial "diff: "
  refute_output --partial "index 33c3d8b"
}
```

Between bats and bats-assert there is indended to be support for something like `assert_line --index 7` to define specific lines of output`, however I was unable to make that work. I believe its a bug between [empty text lines](https://github.com/sstephenson/bats/pull/93) and the escape codes. 

## CI
This is setup with CircleCI, which works great. The Travis bots are crazy slow for me (though it [does work](https://travis-ci.org/paulirish/diff-so-fancy/builds)), so I'd prefer being on CircleCI for now.
image
https://circleci.com/gh/paulirish/diff-so-fancy

Fixes #32

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,6 @@
+[submodule "test/test_helper/bats-core"]
+	path = test/test_helper/bats-core
+	url = https://github.com/ztombol/bats-core
+[submodule "test/test_helper/bats-assert"]
+	path = test/test_helper/bats-assert
+	url = https://github.com/ztombol/bats-assert
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,7 @@
+language: bash
+before_install:
+  - git clone --depth 1 https://github.com/sstephenson/bats.git
+  - sudo apt-get update  -qq
+  - sudo apt-get install -y coreutils
+script:
+  - ./bats/bin/bats test/*.bats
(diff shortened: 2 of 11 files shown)


stevemao:
:tada: Thanks @paulirish , This is so great!
Do I need to enable the hook to make CI work?

paulirish:
@stevemao no you just need to signup with CircleCI and the permissions you grant them will take care of it.

paulirish:
i'd be a fan of merging this earlier than later. shall we?

PR #54, opened 2016-02-13
travis: test on both osx and linux

stevemao:
~~*I'm playing with the CI so don't merge :)*~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,10 @@
 language: bash
 before_install:
   - git clone --depth 1 https://github.com/sstephenson/bats.git
-  - sudo apt-get update  -qq
-  - sudo apt-get install -y coreutils
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update         ; fi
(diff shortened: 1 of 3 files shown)


stevemao:
In fact, this can probably be merged :) but I wanna run it pass @paulirish :)

.travis.yml
```diff
@@ -1,7 +1,10 @@
 language: bash
 before_install:
   - git clone --depth 1 https://github.com/sstephenson/bats.git
-  - sudo apt-get update  -qq
-  - sudo apt-get install -y coreutils
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update         ; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install gnu-sed; fi
```
paulirish: if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; brew install gnu-sed; fi

paulirish:
seems cool. I'll merge and one-line that change.

PR #55, opened 2016-02-13

pauley-unsaturated:
- Fixing problems with symlinked diff-so-fancy

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -9,24 +9,25 @@ hash gsed 2> /dev/null && SED=gsed || SED=sed
 if hash diff-highlight 2> /dev/null; then
 	diff_highlight=diff-highlight
 else
-	SCRIPT_PATH=$( cd $(dirname $0) ; pwd -P )
+	SCRIPT_PATH=$( cd $(dirname $(readlink $0 || echo $0)) ; pwd -P )
 	diff_highlight="$SCRIPT_PATH/third_party/diff-highlight/diff-highlight"
 fi
 
-color_code_regex="(\x1B\[([0-9]{1,3}(;[0-9]{1,3}){0,3})[m|K])?"
-reset_color="\x1B\[m"
-dim_magenta="\x1B\[38;05;146m"
-invert_color="\x1B\[7m"
+CSI=$'\x1b\['
+color_code_regex="(($CSI([0-9]{1,3}(;[0-9]{1,3}){0,3})[m|K])?)"
+reset_color=$'\x1b\[0?m'
+dim_magenta=$'\x1b\[38;05;146m'
+invert_color=$'\x1b\[7m'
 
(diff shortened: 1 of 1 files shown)


diff-so-fancy
```diff
@@ -9,24 +9,25 @@ hash gsed 2> /dev/null && SED=gsed || SED=sed
...
+invert_color=$'\x1b\[7m'
 
 git_index_line_pattern="^($color_code_regex)index .*"
 
 format_diff_header () {
-	# simplify the unified patch diff header
-	$SED -E "s/^($color_code_regex)diff --git .*$//g" | \
-	$SED -E "/$git_index_line_pattern/{N;s/$git_index_line_pattern\n//}" | \
+    # simplify the unified patch diff header
+	$SED -E -e "s/^($color_code_regex)diff --git .*$//" | \
```
pauley-unsaturated: The -e's aren't really needed here, though I prefer to use -e when possible.

diff-so-fancy
```diff
@@ -9,24 +9,25 @@ hash gsed 2> /dev/null && SED=gsed || SED=sed
...
-	SCRIPT_PATH=$( cd $(dirname $0) ; pwd -P )
+	SCRIPT_PATH=$( cd $(dirname $(readlink $0 || echo $0)) ; pwd -P )
 	diff_highlight="$SCRIPT_PATH/third_party/diff-highlight/diff-highlight"
 fi
 
-color_code_regex="(\x1B\[([0-9]{1,3}(;[0-9]{1,3}){0,3})[m|K])?"
-reset_color="\x1B\[m"
-dim_magenta="\x1B\[38;05;146m"
-invert_color="\x1B\[7m"
+CSI=$'\x1b\['
```
pauley-unsaturated: This was the only way I could find to get bsd sed to match escaped ANSI color.  Need '$' + single-quote string.

stevemao:
@pauley-unsaturated tests fail incase your not aware :)

paulirish:
I'm seeing the literal string "?m" in the output. I see this with both gnu sed and bsd sed. 

you can repro with 

    cat test/fixtures/ls-function.diff  | ./diff-so-fancy

image

pauley-unsaturated:
My bad! I'm sure I messed the escape code matching up. I'll try this again.


-Pauley

paulirish:
closed?

i do hope you give this another shot. :)

pauley-unsaturated:
I am giving it a shot right now, but I am doing nasty things to my git history to make it look like I know what I'm doing ;-)

pauley-unsaturated:
Ok, found the issue.  I'll re-issue a PR.

stevemao:
@pauley-unsaturated You don't need to close and reopen PRs. Keep working and finally squash your commits :)

PR #56, opened 2016-02-13
2 1-digit ansi instead of 1 3-digit for dim magenta

### LINKED ISSUE #17: Incorrectly highlighted lines on Windows
I've set up the tool using the instructions from the README (npm install, then set core.pager value). I've run a test diff and the *first code line of each file* is incorrectly highlighted and colored:

capture

Also, the headers are white instead of yellow, but that's a minor issue.

My environment: Windows 10, Cmder console (but issue also occurs in PowerShell), default Git for Windows installer.

stevemao:
Hopefully this is more supported.
Fixes #17

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -15,7 +15,7 @@ fi
 
 color_code_regex="(\x1B\[([0-9]{1,3}(;[0-9]{1,3}){0,3})[m|K])?"
 reset_color="\x1B\[m"
-dim_magenta="\x1B\[38;05;146m"
+dim_magenta="\x1B\[35m\x1B\[2m"
 invert_color="\x1B\[7m"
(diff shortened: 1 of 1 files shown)


paulirish:
Yeah I like this. works for me

paulirish:
It is a little weird to have the foreground color totally match the diff annotation though:
image

Can we tweak the foreground in another way?

stevemao:
That's because iterm doesn't support "dim" (`\x1B\[2m`) (it's supported in the nightly). Open the default terminal you'll see it.

paulirish:
chill:

image


🍦

tleunen:
Not sure the color is great... We can barely see what's written :/

PR #59, opened 2016-02-14

pauley-unsaturated:
- Accepting either CSI0m or CSIm as reset sequences
  - Emitting only CSI0m


pauley-unsaturated:
AppVeyor may have failed due to history rewrite in my branch.  Let's try this one more time.  Sorry for the noise!

stevemao:
Oh don't worry about appveyor, I enabled it but never configured :)

PR #60, opened 2016-02-14

pauley-unsaturated:
- Accepting either CSI0m or CSIm as reset sequences
  - Emitting only CSIm


pauley-unsaturated:
Ok, this PR should have been clean... I have no idea why the windows unit test would fail :(

stevemao:
I'll configure the CI now.

paulirish:
Now that we merged the style PR, we have conflicts. @stevemao can you resolve these manually for pauley?

PR #62, opened 2016-02-14
Strip empty spaces alternative

noscript:
Alternative implementation for #51 which takes into account empty lines.

Before:
before

After:
after

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -55,7 +55,11 @@ mark_empty_lines () {
 
 strip_leading_symbols () {
   # strip the + and -
-  $SED -E "s/^$color_code_regex[\+\-]/\1 /g"
+  $SED -E "s/^($color_code_regex)[\+\-]/\1 /g"
+}
+
+strip_first_column () {
+  $SED -E "s/^($color_code_regex)[[:space:]]/\1/g"
 }
 
 print_horizontal_rule () {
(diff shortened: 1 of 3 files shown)


scottchiefbaker:
I came here to see if this was the same "bug" I'm experiencing, but i can't tell from the screenshots. What is different between the two screenshots?

noscript:
The first whitespace column is stripped away.

scottchiefbaker:
Oh! Now I see it. This is an annoying "issue", I hope this gets merged.

paulirish:
i'd like to proceed with this change. while I have a preference towards indentation, i think the net benefits of dropping the spaces are significant.

@noscript do you want to update the PR?

noscript:
Great to hear. Will update during the day.

noscript:
Ready for merge.

scottchiefbaker:
I use tabs for indent, and with this change anything that's indented one level is exactly one space from the left. Second level indents appear correct though.

indent

The rats line has one level of tab indent, and the resellers line has two levels of tab indentation.

noscript:
Update less flags:
```diff
-less --tabs=1,5 -RFX
+less --tabs=4 -RFX
```

scottchiefbaker:
Oh good call... that fixed it

noscript:
I think we will be seeing few more bug reports about indentation in coming days :)

stevemao:
@so-fancy/diff-so-fancy If you like to make a release please do so. Or I'll do it in a few hours. Thanks.

noscript:
Please do.

xixixao:
@noscript  :clap:

xixixao:
The README's screenshot could be updated to reflect the change.

PR #63, opened 2016-02-14
Shell cleanup

bdossantos:
Hello,

I've added some small style improvement to `./diff-so-fancy` shell script according [Shell Style Guide](https://google.github.io/styleguide/shell.xml) and [Shellcheck](https://github.com/koalaman/shellcheck)

PR #64, opened 2016-02-14
test: add shellcheck to Travis

bdossantos:
## Description

This PR add small tweak to install Shellcheck on Travis and run `shellcheck diff-so-fancy update-deps.sh`

## Depend

#63


paulirish:
This is great. I've used shellcheck before and am really impressed with its feedback.

We'll wait for the first PR to land, obv.

PR #67, opened 2016-02-16
updates Mac install guide for brew; closes #21

### LINKED ISSUE #21: Create and submit a brew formula
Repo more than meets the star/fork requirements. Most Mac users would probably prefer a pure-brew install than a brew+npm install.

`brew install gnu-sed --with-default-names && npm install -g diff-so-fancy` would become `brew install diff-so-fancy`. You could even include the `git config`uration in the formula.

OJFord:


stevemao:
LGTM, gonna merge it and we can always tweak the wording. Thanks @OJFord again!

PR #74, opened 2016-02-16
add test for diff-highlight path resolution

OJFord:
The actual test is identical to the previous check that highlighting is working. I can't see a way of re-running a test with a different `setup`/`teardown`?


paulirish:
@OJFord oh this is awesome. 

Let's move it into its own file: `diff-highlight.bats`, just to keep its setup/teardown separate from the rest.

paulirish:
thx!

PR #75, opened 2016-02-16
Support BSD sed

### LINKED ISSUE #72: Horizontal rules not displaying correctly.
Please see the image. I'm using the Solarized Dark xterm-256color theme from https://github.com/mathiasbynens/dotfiles

diff-so-fancy

elyscape:
Fixes: #72


pauley-unsaturated:
This is the same as my change, and it works for me.

Issue #26, opened 2016-02-10
Can't use the arrow keys to scroll up/down during the diff


stringbeans:
Not sure if this is even fixable since you're piping the output but it would be nice :)

stevemao:
Doesn't happen on my machine. I believe it's related to your [`core.pager`](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration)

chubas:
Same issue, I lose paging capabilities when using it. Workaround is to add a ` | less -R` on the git alias.
My pager is set to `less -FRSX` which seems to be pretty standard.

stevemao:
@chubas @stringbeans The readme already mentions `git config --global core.pager "diff-so-fancy | less --tabs=1,5 -R"`, this should work?

gfpacheco:
I fixed it by removing the `X` from the less flags on the instructions:

    git config --global pager.diff "diff-so-fancy | less --tabs=1,5 -RF"
    git config --global pager.show "diff-so-fancy | less --tabs=1,5 -RF"

PR #76, opened 2016-02-16
Remove -RFX flags from less; git provides these

elyscape:
[Git will by default set the `LESS` environment variable to `FRX`][git-less], so we don't need to set it manually.

[git-less]: https://github.com/git/git/blob/v2.7.1/Documentation/config.txt#L646-L648

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -21,8 +21,8 @@ git diff --color | diff-so-fancy
 
 **But**, you'll probably want to fancify all your diffs. Run this so `git diff` (and `git show`) will use it:
 ```shell
-git config --global pager.diff "diff-so-fancy | less --tabs=1,5 -RFX"
-git config --global pager.show "diff-so-fancy | less --tabs=1,5 -RFX"
+git config --global pager.diff "diff-so-fancy | less --tabs=1,5"
+git config --global pager.show "diff-so-fancy | less --tabs=1,5"
 ```
 
 However, if you'd prefer to do the fanciness on-demand with `git dsf`, drop an alias in your `~/.gitconfig`:


stevemao:
I don't have the problem since I didn't change the defaults. But for people who have might need those flags.

elyscape:
@stevemao The flags are provided by Git in the `LESS` environment variable.

stevemao:
@elyscape https://github.com/so-fancy/diff-so-fancy/issues/26 and https://github.com/so-fancy/diff-so-fancy/pull/50

elyscape:
Interesting. I wonder if that's a result of using a different version of git.

paulirish:
@arthurdarcet can you chime in?

elyscape:
Looking at the documentation history for git, [they were passing in `FRSX` at least as early as 1.8.5](https://github.com/git/git/blob/v1.8.5/Documentation/config.txt#L562-L569). I can't speak to earlier versions like 1.8.3, though, which I know is used on EL7, or 1.7.x, which was used on EL6.

arthurdarcet:
I actually have a `LESS = -R` in my shell on OSX. I can't where it comes from so I can't be sure, but it might be added by OSX directly which is why I needed those flags added explicitly.

Anybody with another mac and a ~stock shell could confirm the `LESS` env variable is set by the OS to `-R` by default? (with `echo $LESS`)

elyscape:
OS X over here and `LESS` is unset.

arthurdarcet:
my bad, found it: https://github.com/robbyrussell/oh-my-zsh/blob/master/lib/misc.zsh#L23

so +1 on this PR

elyscape:
@stringbeans, could you take a look at your environment?

stevemao:
I feel there is no harm keeping the flags as people might change the default ones.

paulirish:
> I feel there is no harm keeping the flags as people might change the default ones.

I agree. The git docs say that Git will apply their flags only if someone doesn't have their own `LESS` env defined. If they do, Git does nothing. 

Our current config guarantees we wont have problems. Yes it's a tad uglier, but it's easier to maintain for people's various configurations.

PR #83, opened 2016-02-18
rename perl interpreter to /usr/bin/env perl

matejc:
I and other Nix/NixOS users do not have perl at location /usr/bin/perl


stevemao:
Hi @matejc , Thanks for the PR. But we don't maintain `diff-highlight`, when we run `update-deps.sh` it would be overwritten. Send a PR to https://github.com/git/git/tree/master/contrib :)

PR #87, opened 2016-02-21
fix #86: update alias example

### LINKED ISSUE #86: git simple alias with pipes and multiple arguments doesn't work.
e.g:
```
$> git dsf --cached test/
fatal: ambiguous argument '@$': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

Instead we can use bash function to make the alias work with pipes and all.

rajasharan:


jebaum:
Heads up, I encountered this problem myself and fixed it in the same way, but there's a second issue overlooked here. This alias will (generally) only work when in the root directory of a git repository. If you are in a subdirectory and, naturally, provide a relative path to a file, you will get an error about an "unknown revision or path not in the working tree". This is because of a git misfeature that changes the working directory to the repository root when executing aliases, as documented in the `alias.*` section of the `git-config` man page. To fix this, change the alias to:

`dsf = "!f() { [ \"$GIT_PREFIX\" != \"\" ] && cd "$GIT_PREFIX"; git diff --color $@ | diff-so-fancy | less; }; f"`

paulirish:
@jebaum is the behavior any different if the user does `git diff .`  ?  TBH, that's what i've always done

paulirish:
@rajasharan thanks for the PR. 

however, since we're adding `less` into it, lets use the same less arguments as our default.

stevemao:
@tommarshall Looks good to you?

tommarshall:
The change looks good to me.

However I can verify the additional issue reported by @jebaum:

```
tom at mbp in ~/work/github/ansible-role-mycli/tasks on master*
$ git diff setup-Debian.yml
diff --git a/tasks/setup-Debian.yml b/tasks/setup-Debian.yml
index d2ad443..dc16973 100644
--- a/tasks/setup-Debian.yml
+++ b/tasks/setup-Debian.yml
@@ -9,6 +9,7 @@

 - name: Update apt cache.
   apt: update_cache=yes cache_valid_time=86400
+  tags: tmp

 - name: Install apt-transport-https.
   apt: name=apt-transport-https state=latest
```
```
tom at mbp in ~/work/github/ansible-role-mycli/tasks on master*
$ git dsf setup-Debian.yml
fatal: ambiguous argument 'setup-Debian.yml': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

I didn't know about this "misfeature" of git aliases.

Not sure if you want to handle that separately or not, but @jebaum's fix together with the default `less` arguments is working for me:

```
dsf = "!f() { [ \"$GIT_PREFIX\" != \"\" ] && cd "$GIT_PREFIX"; git diff --color $@ | diff-so-fancy | less --tabs=1,5 -RFX; }; f"
```

paulirish:
I've updated the readme with the new alias. thx yall!

PR #90, opened 2016-02-24
Clean up the headers to indicate a file was added/removed/modified

scottchiefbaker:
Also makes the file name "clickable" by removing the leading a/ and /b

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -70,6 +70,10 @@ print_horizontal_rule () {
   printf "%$(tput cols)s\n"| $SED 's/ /─/g'
 }
 
+print_header_clean () {
+  "$(get_script_dir)/libs/header_clean/header_clean.pl"
+}
+
 # run it.
 # shellcheck disable=SC2002
 cat $input \
@@ -79,4 +83,5 @@ cat $input \
   | mark_empty_lines \
   | remove_first_empty_line \
   | strip_leading_symbols \
-  | strip_first_column
+  | strip_first_column \
+  | print_header_clean
--- a/libs/header_clean/header_clean.pl
+++ b/libs/header_clean/header_clean.pl
@@ -1,0 +1,43 @@
+#!/usr/bin/perl
+
+use strict;
+use warnings;
+
+my $ansi_sequence_regex = qr/(\e\[([0-9]{1,3}(;[0-9]{1,3}){0,3})[mK])?/;
+
+my ($f1,$f2);
+while (my $l = <>) {
+	# Find the first file: --- a/README.md
+	if ($l =~ /^$ansi_sequence_regex--- (a\/)?(.+?)(\e|$)/) {
+		$f1 = $5;
+	# Find the second file: +++ b/README.md
+	} elsif ($l =~ /^$ansi_sequence_regex\+\+\+ (b\/)?(.+?)(\e|$)/) {
+		if ($1) {
+			print $1; # Print out whatever color we're using
+		}
+		$f2 = $5;
+
+		# If they're the same it's a modify
+		if ($f1 eq $f2) {
+			print "modified: $f1\n";
+		# If the first is /dev/null it's a new file
+		} elsif ($f1 eq "/dev/null") {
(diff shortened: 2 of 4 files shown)


scottchiefbaker:
My sed skills are seriously lacking, but Perl skills are decent. I went ahead and took a stab at addressing #28 but calling a Perl script I put in `/third_party/`

Let me know if I'm going down the wrong path here.

paulirish:
i believe this would qualify as first-party code, unless you're releasing this separately. :)

I hesitate a bit to :+1: adding perl to our sourcebase. I'm certainly way less comfortable in it than bash.  Would you consider porting your work here to bash?

paulirish:
curious what others think, too. I'm flexible if the consensus leans toward the perl implementation.

stevemao:
I'm concerned with how we can handle one extra dependency... (I miss node :smile:) We need to update the install section in README and might need to deal with related issues, etc.

noscript:
`diff-highlight` is a Perl script, so shouldn't be an issue.

paulirish:
yeah install-wise it wont be a problem, so that's an upside.

@scottchiefbaker wanna write a few tests?

noscript:
As `diff-so-fancy` gets more complicated, I would rather rewrite it entirely in Perl. Which would also make it faster.

scottchiefbaker:
I don't know enough bash to make this sort of thing. Someone with more bash skills wouldn't have too much of a problem, the script is pretty simple. 

If we're comfortable with a Perl implementation I can take a look at adding tests. Otherwise I'm happy to work with someone who wants to write a bash implementation, to explain my logic.

paulirish:
yeah i am feeling the pain of bash as well. i'm into porting everything into perl. 
(better yet, javascript, but that's a dependency that may be too much to ask)

scottchiefbaker:
I know enough Perl that I can contribute more. What's the next/first step to going down that road? Do we want to land this change for clickable headers and start a separate issue for migrating some of the other pieces to Perl?

paulirish:
> As diff-so-fancy gets more complicated, I would rather rewrite it entirely in Perl. Which would also make it faster.

I was curious about this so I ran some numbers.

### perf of diff-so-fancy

I took a fairly busy diff of ~10 files as my example. I broke apart the 4 pieces of the pipeline, and ran them a few times. These are the rough medians:

```
      git diff: █████ 50ms
diff-highlight: ██████████ 100ms
 diff-so-fancy: █████████████ 130ms
  header_clean: ███ 30ms
```

(I was running them with `time bash -c "…"` so there's ~10-20ms of overhead built into the numbers.)

Pretty interesting. Bash is certainly not giving us speed here. :)

noscript:
Multiple calls to `sed` is the major slow down. We could make the current version faster if do all in one call, but the readability will suffer.

scottchiefbaker:
I count 9 calls to sed in diff-so-fancy. At least four of them (in `format_diff_header`) could easily be combined.

paulirish:
I'm +1 for merging this.  

I'm also interested in migrating the rest of the script into perl, as I think maintenance and readability would improve.

@scottchiefbaker can you move the script to a new `./lib` folder? 

Also. a test? :)

scottchiefbaker:
Yes I can move this to a `.libs` folder.

I looked at adding tests, and I can't even get the existing tests to run locally. No matter what I do I get: 

    bats: /home/bakers/github/diff-so-fancy/test/test_helper/bats-core/load.bash does not exist

Any tips on setting up a testing environ so I can start writing tests?

paulirish:
@scottchiefbaker did you `git submodule update --init ` ?

scottchiefbaker:
I cleaned up my pull request and added tests for the functionality I added.

paulirish:
lookin sharp. merging…

stevemao:
@scottchiefbaker can you have a look at https://github.com/Homebrew/homebrew/pull/49712? Thanks.

scottchiefbaker:
Checking it out now...

PR #91, opened 2016-02-24
Delete whitespace above header line (instead of replacing it with "")

scottchiefbaker:
Before:
before

After:
after

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -38,7 +38,7 @@ git_index_line_pattern="^($color_code_regex)index .*"
 
 format_diff_header () {
   # simplify the unified patch diff header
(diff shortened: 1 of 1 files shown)


paulirish:
After merging this, I realize that I had it there deliberately. I prefer the extra space especially between changes..

This feels too dense to me: 
image

I'm going to put up a PR that  adds a newline after each file, which I think will make both of us happy. :)

scottchiefbaker:
Oh really? I prefer it the way it is, I think the start yellow is a good divider. I'm fine either way, I was most concerned about the line before the diff started, as that was 100% wasted space. As long as that line is gone I'm fine.

PR #92, opened 2016-02-26
add some newlines between files for spacing and readability

paulirish:
Screenshot of before and after:
image

wdyt, @scottchiefbaker ?

here's a sample without the pager:
image


scottchiefbaker:
Works for me... but it is another call to `sed` which may affect performance. I'm not overly concerned with performance though.

paulirish:
yup. We'll get the perf bit sorted out soon

PR #95, opened 2016-02-29
add fixes for powershell

lamchau:
* 4e5800f; Related to #7
   * I ran into this issue on a Windows console emulator (ConEmu; cmder) using a bitmap font (ProggyCleanTTSZ). 
```
ΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓö

--- a/fish/functions/ls.fish
+++ b/fish/functions/ls.fish

ΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓö

@@ -7,6 +7,8 @@
```

 * cb0c0f1
   * For some reason PowerShell wraps a line at terminal width and needs a small tweak that keeps the header formatted properly.

-----
**Fixes tested on Windows (PowerShell, cygwin), OS X (iTerm, terminator, terminal)**

**Windows (cmder)** (via `npm link`)
image

**OSX (terminal)**
image

### DIFF
--- a/test/diff-so-fancy.bats
+++ b/test/diff-so-fancy.bats
@@ -22,7 +22,7 @@ if begin[m"
   refute_output --partial "index 33c3d8b..fd54db2 100644"
 }
 
-@test "+/- line stars are stripped" {
+@test "+/- line symbols are stripped" {
   refute_output --partial "
 [1;31m-"
   refute_output --partial "
--- a/test/diff-so-fancy.bats
+++ b/test/diff-so-fancy.bats
@@ -43,3 +43,11 @@ if begin[m"
   output=$( load_fixture "noprefix" | $diff_so_fancy )
   refute_output --partial "diff --git setup-a-new-machine.sh"
 }
+
+@test "header format uses a native line-drawing character" {
+  header=$( printf "%s" "$output" | head -n3 )
+  run printf "%s" "$header"
+  assert_line --index 0 --partial "[1;33m─────"
+  assert_line --index 1 --partial "modified: fish/functions/ls.fish"
+  assert_line --index 2 --partial "[1;33m─────"
+}
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -64,6 +64,10 @@ strip_first_column () {
 print_horizontal_rule () {
   let width="$(tput cols)"
 
+  if [[ $(uname -s) =~ (MINGW32*|MSYS*) ]]; then
+    width=$(( width - 1 ))
+  fi
+
   # echo -n '─' | hexdump -C
   local -r dash=$( printf "%b" "\xe2\x94\x80" )
   printf "%*s\n" "$width" | $SED "s/ /${dash}/g"
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -62,7 +62,11 @@ strip_first_column () {
 }
 
 print_horizontal_rule () {
-  printf "%$(tput cols)s\n"| $SED 's/ /─/g'
+  let width="$(tput cols)"
+
+  # echo -n '─' | hexdump -C
+  local -r dash=$( printf "%b" "\xe2\x94\x80" )
+  printf "%*s\n" "$width" | $SED "s/ /${dash}/g"
 }
 
 print_header_clean () {


stevemao:
Would be great if you could add a test for the fix :) @lamchau

lamchau:
@stevemao sounds good :) working on it!

paulirish:
Yup. let's add a test. You'll do a `refute_output --partial ` to check for newline followed by "modified" 
This modified text was just added to master a few minutes ago.. so you'll have to merge locally. :)

lamchau:
`bats` is giving me false positives and i can't seem to validate the line-drawing character exists

```sh
@test "header uses native line-drawing character" {
  # using assert to validate the character shows up (i've tried pasting '─')
  assert_output --partial "[1;33m(0q(B"
}
```

any ideas?

paulirish:
i'd go with

    assert_output --partial "[1;33m──────"

... looks like github doesn't really like that sequence. 

But I generate it like so:

    git --no-pager diff | diff-so-fancy > output.txt

and use that to drop into the tests.

lamchau:
@paulirish ahh, yeah i dumped the entire header out and found it still wouldn't pass. thanks for the fix!

ps: i split the line into into 2 tests to ensure both a start/end escape sequence exist -- did we want tests for the `uname` fix? travis also seems to be taking awhile to pick up the changes

paulirish:
@lamchau yeah we want a test for the line-length issue that you're sniffing via uname.

test/diff-so-fancy.bats
```diff
@@ -40,3 +40,8 @@ if begin[m"
   assert_output --partial 'eval [m[1;31;48;5;52m"env CLICOLOR_FORCE=1 command $ls $param [m[1;31m$argv"[m'
   assert_output --partial 'eval [m[1;32;48;5;22m$ls $param "[m[1;32m$argv"[m'
 }
+
+@test "header uses native line-drawing character" {
+  assert_output --partial "(0q"
```
paulirish: you'll want to capture more, so we dont catch a false positive:

```sh
  assert_output --partial "
[1;33m(0qqqqq"
```

test/diff-so-fancy.bats
```diff
@@ -40,3 +40,8 @@ if begin[m"
...
+  assert_output --partial "(B"
```
paulirish: I wonder if you can verify the linewrap isnt broken in a similar way:

```sh
assert_output --partial "qqq(B
[1;33mmodified"
```

basically that there's no extra line break between the end of the line and the file header. give it a whirl?

test/diff-so-fancy.bats
lamchau: fixed

test/diff-so-fancy.bats
lamchau: ah that make sense, yeah

test/diff-so-fancy.bats
paulirish: did you push?

lamchau: let me push now, i'm still trying to figure out how to "mock" `uname`

lamchau:
my attempt so far (but running into subshell issues)

```sh
@test "header length has the correct length" {
  #setup
  function mock() {
    echo "$1"
  }
  shopt -s expand_aliases
  alias actual_tput=$(tput cols)
  alias tput="mock '80'"
  alias uname="mock 'Linux'"

  # test
  local first_line=$( load_fixture "ls-function" | $diff_so_fancy | head -n1 )
  [[ ${#first_line} -eq 80 ]]

  alias uname="mock 'MSYS_NT-6.1-WOW'"
  first_line=$( load_fixture "ls-function" | $diff_so_fancy | head -n1 )
  [[ ${#first_line} -eq 79 ]]

  # teardown
  unset -f mock
  unalias tput
  unalias uname
  shopt -s expand_aliases
}
```

paulirish:
Hmmm. I'm curious if we can try something simpler... 

1. grab the second line of output
1. assert it doesn't have any drawing character
1. assert it has the `modified: fish/functions/ls.fish` text.

would that fail before your patch? also hypothetically it wouldn't need any mocking.  

Thanks for working on this! pretty excited.

lamchau:
yeah, let's just do the simpler test. we're compensating for a (probable) bug in powershell anyways and i couldn't get `bats` to run the test suite in powershell to truly validate 

good catch with the newline though, the change was rebased away but i ended up having to move the newline
```diff
- printf "%s%s%s\n" $'\e(0' "$separator" $'\e(B'
+ printf "%s%s%s" $'\e(0' "$separator" $'\e(B\n'
```

tests are covered in c10206d and should be sufficient -- what do you think?

paulirish:
oh shit! you got ` assert_line --index 0` to work!  I was failed at that completely when i had tried. :)


lookin real sharp. 
we might eventually just do `width - 1` for everyone, because why not. but this is solid for now.

thanks for your work here!

Issue #100, opened 2016-03-01
Should we change how hunk headers look?


scottchiefbaker:
Raw diffs include the lines numbers encoded strangely:

![lines](https://cloud.githubusercontent.com/assets/3429760/13437544/9e269ef0-df99-11e5-8e6b-94719cb13985.png)

Note the magenta line numbers surrounded by @@. Should we:

1. Get rid of them completely
2. Change them to "* Line #XX"
3. Something totally different
4. Leave them alone

lode:
Yeah! More readability can be added there for sure.

As I decompile it, it is the line number where the diff starts (including context lines), and the amount of lines.

For myself, I'm mostly looking for the line start, but then also of the first changed line, so I can enter it in my editor and go there, exactly there. This is (as I understand), the third number + context lines. Something alike `:42` or `:42 @ function foo()` where 42 is the line of the first changed/added/removed line. (The syntax is taken from how sublime uses this in the goto symbol/line dialog.)

But I guess other people use the presented data for other purposes?

scottchiefbaker:
I reworked my script to capture the @@ lines and display them differently, I'm just not sure what "differently" looks like right now. I started off with just: "* Line #3 *" as that's raw from the diff. 

I use it the same way you do, "where do I go after I open up the this file". Maybe it should be the line number + the context lines (i.e. go directly to the first changed line)?

scottchiefbaker:
This is an example of what I've come up with so far:

```
───────────────────────────────────────────────────────────────────────────────────────
modified: include/version.inc
───────────────────────────────────────────────────────────────────────────────────────
* Changes start on line #22 *
 */

/* PHP 7.0 Release */
$PHP_7_0_RC = false; // Current RC version (e.g., '5.6.7RC1') or false
$PHP_7_0_RC_DATE = '21 Jan 2016';
$PHP_7_0_RC = "7.0.4RC1"; // Current RC version (e.g., '5.6.7RC1') or false
$PHP_7_0_RC_DATE = '18 Feb 2016';

$PHP_7_0_VERSION         = "7.0.3";
$PHP_7_0_DATE            = "04 Feb 2016";
* Changes start on line #39 * $PHP_7_0_SHA256     = array(
);

/* PHP 5.6 Release */
$PHP_5_6_RC = false; // Current RC version (e.g., '5.6.7RC1') or false
$PHP_5_6_RC_DATE = '21 Jan 2016';
[...]

PR #102, opened 2016-03-03
Rework file permission headers

scottchiefbaker:
This addresses #100

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -38,8 +38,7 @@ git_index_line_pattern="^($color_code_regex)index .*"
 
 format_diff_header () {
   # simplify the unified patch diff header
-  $SED -E "s/^($color_code_regex)diff --git .*$//g" \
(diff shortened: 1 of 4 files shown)


paulirish:
I'm seeing the `diff --git` line in my output with this patch:
image

scottchiefbaker:
Interesting... your diff output doesn't have `a/.bashrc` or `b/.bashrc` on it like mine does. I just updated the pull request to handle both types.

paulirish:
@scottchiefbaker true. once i found `git config --global diff.noprefix true` i've been loving it. :)  But ideally i don't need it when using diff-so-fancy.

paulirish:
@scottchiefbaker as you can see i just added a test for that one. :)

paulirish:
ah so did you. but i added a fixture that starts with noprefix output. so between the two, we're good. ;)

PR #104, opened 2016-03-03
update bats-support test deps

paulirish:
bringing these up to date with latest https://github.com/ztombol/bats-support


paulirish:
accidentally pushed a branch to this repo, instead of my user account. then i deleted the remote branch. CI got confused. but it should be fine.

PR #106, opened 2016-03-04
update testing readme

lamchau:
prevents `fatal: no submodule mapping found in .gitmodules for path 'test/test_helper/bats-core'`


paulirish:
ah yeah since i moved things. thx

PR #107, opened 2016-03-04
Rework how git hunks are displayed

scottchiefbaker:
This addresses things discussed in #100 

Also there is a minor bugfix in a92e9cf. If you ran diff-so-fancy with no changed files there would have been a warning emitted because there was not output to parse.

### DIFF
--- a/libs/header_clean/header_clean.pl
+++ b/libs/header_clean/header_clean.pl
@@ -109,7 +109,7 @@ sub strip_empty_first_line {
 	my $foo = shift(); # Array passed in by reference
 
 	# If the first line is just whitespace remove it
-	if ($foo->[0] =~ /^\s*$/) {
+	if (defined($foo->[0]) && $foo->[0] =~ /^\s*$/) {
 		shift($foo);
 	}
 }
--- a/libs/header_clean/header_clean.pl
+++ b/libs/header_clean/header_clean.pl
@@ -2,10 +2,12 @@
 
 use strict;
 use warnings;
+use File::Basename;
 
 my $remove_file_add_header    = 1;
 my $remove_file_delete_header = 1;
 my $clean_permission_changes  = 1;
+my $change_hunk_indicators    = 1;
 
 #################################################################################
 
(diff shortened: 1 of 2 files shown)


scottchiefbaker:
Sample of what my changes look like

one1

scottchiefbaker:
Also of note... If the diff starts on the very first, or very last line the diff hunk indicator will be off by 2 or 3 lines. The diff hunk includes the lines the changes BEGIN on, which includes three lines of context. I compensate for that here: 

https://github.com/scottchiefbaker/diff-so-fancy/blob/diff_hunks/libs/header_clean/header_clean.pl#L76

If the hunk starts at the beginning or end there won't be context and my math will be off by a little. I thought for a long time and couldn't think of a way to compensate for that. If someone else has a better idea I'm open to suggestions.

paulirish:
Niiiice! this feels great. :D 

Small bug here.... doubling up the filename. might be my noprefix hitting again:
image

scottchiefbaker:
@paulirish can you send me the raw `git diff` output that generated that? I'll see if I can't figure out the cause and update the pull request.

paulirish:
```
[1;33mdiff --git .bash_prompt .bash_prompt[m
[1;33mindex 852d69f..8d3e3d0 100644[m
[1;33m--- .bash_prompt[m
[1;33m+++ .bash_prompt[m
[1;35m@@ -4,6 +4,9 @@[m
 default_username='paulirish'[m
 [m
 [m
[1;32m+[m[1;32meval "$(thefuck --alias)"[m
[1;32m+[m
[1;32m+[m
 if [[ -n "$ZSH_VERSION" ]]; then  # quit now if in zsh[m
     return 1 2> /dev/null || exit 1;[m
 fi;[m
```

paulirish:
(sorry everyone for always using my very random dotfiles edits in my examples. it's just the messiest diff scene i have. :)

scottchiefbaker:
You were right, it was the noprefix thing. I updated my code to properly handle both cases.

paulirish:
lgtm!

scottchiefbaker:
:+1:

bpinto:
@paulirish

Why does it come with these colors characters? `�[1;35m` (I believe they are color characters).
Is there a way to avoid this?

scottchiefbaker:
@bpinto `\e[1;35m` is bright/bold purple. Is that a problem?

bpinto:
I'm using `pbcopy` and trying to share it on slack but this color "text" comes with it making it harder to read the changes.

stevemao:
@bpinto have a look at https://github.com/so-fancy/diff-so-fancy#reporting-bugs

PR #108, opened 2016-03-05
use .gitconfig for configuration

### LINKED ISSUE #69: Option to enable pesky + or - at line-start
Some people like myself might actually prefer to have the `+` and `-` on the left but still benefit from the other improvements. Feasible?

lamchau:
this is just a proof of concept/work in progress on how we can add user-defined configurations. i chose this route over arguments so you can configure this on the `--local` level of `--global` (or `--system`) 

this is an example for #69 (`--plus-signs`). it would allow for easier customizations in the future

points of interest/questions:
 * git only accepts alphanumeric keys hence `diffsymbols` vs something more readable
 * configurations can grow to have a number of permutations, how exactly do we want to test this?
 * **help**: i've been running into issues with `header_clean`
   1. [on master] `diff --git a/diff-so-fancy b/diff-so-fancy` is always visible
   2. <strike>[on this branch] doesn't want to take input from `cat`</strike> fixed via 2549765
 * tbd: how we define the sections (see below)

**option 1**: shared namespace/section (tidier and more idiomatic)
```sh
[diff-so-fancy]
  diffsymbols = false
```

**option 2**: namespace by feature/section (potentially busy)
```sh
[diff-so-fancy "lines"]
  diffsymbols = true
```

image

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -15,6 +15,18 @@ get_script_dir () {
   echo "$dir"
 }
 
+get_config() {
+  local -r key="diff-so-fancy.$1"
+  local -r default_value="$2"
+  value="$(git config --get "$key")"
+
+  if [[ -z "$value" && ! -z "$default_value" ]]; then
+    value="$default_value"
+  fi
+  printf "%s" "$value"
+}
+
+
 # find my sed
 hash gsed 2> /dev/null && SED=gsed || SED=sed
 
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -47,6 +47,7 @@ dim_magenta="${CSI}38;05;146m"
 invert_color="${CSI}7m"
 
 git_index_line_pattern="^($color_code_regex)index .*"
+diffsymbols="$(get_config 'diffsymbols' false)"
 
 format_diff_header () {
   # simplify the unified patch diff header
@@ -65,12 +66,20 @@ mark_empty_lines () {
 }
 
 strip_leading_symbols () {
-  # strip the + and -
-  $SED -E "s/^($color_code_regex)[\+\-]/\1 /g"
+  if [[ "$diffsymbols" = true ]]; then
+    cat
+  else
+    # strip the + and -
+    $SED -E "s/^($color_code_regex)[\+\-]/\1 /g"
+  fi
 }
 
 strip_first_column () {
-  $SED -E "s/^($color_code_regex)[[:space:]]/\1/g"
+  if [[ "$diffsymbols" = true ]]; then
+    cat
+  else
+    $SED -E "s/^($color_code_regex)[[:space:]]/\1/g"
+  fi
 }
 
 print_horizontal_rule () {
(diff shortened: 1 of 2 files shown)


paulirish:
This is super cool. :)

I am into this idea and think using gitconfig is a great idea.  
definitely option 1 for the config choices.

testing; we can just test with one way, then apply the config change and repeat some/all tests. 

I think you should keep going here. others, what do you think?

lamchau:
@paulirish guess we'll just need tests and we're good to go? :)
cc: @stevemao what do you think about this approach?

paulirish:
tests: yes

i also feel like `get_config` could probably be more terse.

stevemao:
I think we should still support flags. flags takes highest precedence. each flag should have corresponding git config.

> i also feel like get_config could probably be more terse.

I agree. We might have more configs in the future.

lamchau:
yeah fair point, i'll clean up `get_config` and add tests. i'd rather save the flags for another pr and use the configuration for another feature/issue to keep this cleaner

sound good?

scottchiefbaker:
I really like the idea of this!!! :+1: 

What other options should we allow the user to toggle? I'd like to see "colorize_context_line" as toggleable.

scottchiefbaker:
We talked about moving a lot of the sed logic in to Perl for speed. I was thinking about moving the logic for:

1. Strip leading symbols
1. Strip first column 
1. Colorize context line

in to Perl. Before we land this (these are changes to bash/sed) maybe I should work on this?

paulirish:
> What other options should we allow the user to toggle?

I think this is the complete list of what _could_ be configurable.

1. enable diffsymbols
1. disable context line coloring
1. set context line coloring color
1. disable DSF and but leave diff-highlight on? (so the complete output is patch-friendly)
1. disable diff-highlight (but keep the rest of DSF)
1. disable hunk reformatting
1. disable file header reformatting

lamchau:
not gonna lie, i'm a bit lazy to tackle all of these by myself (namely tests) :) anyone want to help out?

so here's a list of things to do set configuration for
 - [x] enable diffsymbols
 - [ ] disable context line coloring
 - [ ] set context line coloring color
 - [ ] disable DSF and but leave diff-highlight on? (so the complete output is patch-friendly)
 - [ ] disable diff-highlight (but keep the rest of DSF)
 - [ ] disable hunk reformatting
 - [ ] disable file header reformatting

future:
 - [ ] precedence of flags > git config (anyone else have an opinion on this? maybe create an issue)

scottchiefbaker:
I don't want to be a buzzkill, but I just submitted a PR (#123) to move some of the sed functionality in to Perl. This will include a very easy way to turn on/off individual options. Some of that code may supersede some of the stuff you started here.

More importantly I submitted `get_git_config()` as part of my commit which will fetch the whole git config and return it as a hash. Once this lands it will be trivial to implement the stuff you mention here.

paulirish:
Let's resume this work on top of the new perl backend. Scott added some details of it here: https://github.com/so-fancy/diff-so-fancy/pull/123#issuecomment-194976713

Issue #109, opened 2016-03-06
header_clean.pl error: Argument "-1 +0" isn't numeric


paulirish:
Getting this.. Appears to be on certain hunk headers..

> :Argument "-1 +0" isn't numeric in addition (+) at /Users/paulirish/code/diff-so-fancy/libs/header_clean/header_clean.pl line 71, <> line 506.

Looks like it's running up against this input:

```
diff --git a/bin/push b/bin/push
deleted file mode 120000
index 3208f92..0000000
--- a/bin/push
+++ /dev/null
@@ -1 +0,0 @@
-git-friendly/push
\ No newline at end of file
```

in particular ` -1 +0,0 ` when it was expecting a `,X` after the `-1`

paulirish:
Fix incoming…

PR #110, opened 2016-03-06
Change hunk line position: original --> result location

paulirish:
If we take an example hunk header like `@@ -54,8 +51,7 @@`.. I think we want to surface the `51` as the line number as it's the likely currently representation of the change on disk. There are other ways in which diffs are viewed, but I think the most common usecases when linenumber is used by our users will be where the result line number is preferred to the original line number.

This change this also provides the ability to fix the error from #109. 
(However if we wanted to fix that without changing logic, we'd tweak this regex from `/(.+?),/` to `/-(\d+)/`)

PR adds a test for the bug. And it rebaselines the existing tests.

@scottchiefbaker whatcha think?

cc @lode

### DIFF
--- a/libs/header_clean/header_clean.pl
+++ b/libs/header_clean/header_clean.pl
@@ -67,7 +67,7 @@ for (my $i = 0; $i <= $#input; $i++) {
 			print $1; # Print out whatever color we're using
 		}
 
-		my ($start_line) = $file_str =~ m/(.+?),/;
+		my ($start_line) = $file_str =~ m/\+(\d+)/;
 		$start_line      = abs($start_line + 0);
 
 		$last_file_seen = basename($last_file_seen);
--- a/libs/header_clean/header_clean.pl
+++ b/libs/header_clean/header_clean.pl
@@ -59,22 +59,21 @@ for (my $i = 0; $i <= $#input; $i++) {
 	########################################
 	# Check for "@@ -3,41 +3,63 @@" syntax #
 	########################################
-	} elsif ($change_hunk_indicators && $line =~ /^${ansi_sequence_regex}@@@* (.+?) @@@*(.*)/) {
-		my $file_str     = $4;
+	} elsif ($change_hunk_indicators && $line =~ /^${ansi_sequence_regex}(@@@* .+? @@@*)(.*)/) {
+
+		my $hunk_header  = $4;
 		my $remain       = $5;
 
 		if ($1) {
 			print $1; # Print out whatever color we're using
 		}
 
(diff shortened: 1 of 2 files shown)


scottchiefbaker:
I agree, I think this makes sense to me. Does the second header ALWAYS start with a plus sign? I think it could contain either a plus or minus sign. We should account for that also.

libs/header_clean/header_clean.pl
```diff
@@ -67,7 +67,7 @@
 			print $1; # Print out whatever color we're using
 		}
 
-		my ($start_line) = $file_str =~ m/(.+?),/;
+		my ($start_line) = $file_str =~ m/\+(\d+)/;
```
scottchiefbaker: I think maybe this should be:

```
my ($start_line) = $file_str =~ m/ [+-](\d+)/;
```

i.e. "A space followed by a plus or minus sign"

paulirish:
> Does the second header ALWAYS start with a plus sign? 

yup!

here's where it's printed. https://github.com/git/git/blob/ab5d01a29eb7380ceab070f0807c2939849c44bc/diff.c#L740

[also](https://github.com/git/git/blob/ab5d01a29eb7380ceab070f0807c2939849c44bc/diff.c#L564)

```c
	/*
	 * As a hunk header must begin with "@@ -<old>, +<new> @@",
	 * it always is at least 10 bytes long.
	 */
```

and ... look at that... `sub parse_hunk_header` in [/git-add--interactive.perl](https://github.com/git/git/blob/ab5d01a29eb7380ceab070f0807c2939849c44bc/git-add--interactive.perl#L798-L805):

```pl

sub parse_hunk_header {
	my ($line) = @_;
	my ($o_ofs, $o_cnt, $n_ofs, $n_cnt) =
	    $line =~ /^@@ -(\d+)(?:,(\d+))? \+(\d+)(?:,(\d+))? @@/;
	$o_cnt = 1 unless defined $o_cnt;
	$n_cnt = 1 unless defined $n_cnt;
	return ($o_ofs, $o_cnt, $n_ofs, $n_cnt);
}
```

handy. :)

libs/header_clean/header_clean.pl
paulirish: sounds good on adding `\d` to it. 

But based on the new info i commented below, i think we can keep it just plus, and no minus at all.

paulirish:
writing regexes is hard.. lets go shopping!

updated to just use parse_hunk_header.

scottchiefbaker:
`parse_chunk_header()` should be more accurate anyway. I like it.

PR #111, opened 2016-03-06
test suite shouldnt assume diff colors

paulirish:
At first these were required, but they appear to no longer be neccessary. \o/


paulirish:
Made some updates here to reduce complexity. 

Still not done, the CI bots seem to be outputting rather different  color codes, and currently the error is in the "does diff-highlight still work even with a weird PATH setup" test.

I'll keep working on this.

scottchiefbaker:
Looks like you broke ALL TEH TESTS.

paulirish:
Yah. Still not there.

PR #116, opened 2016-03-08
Clean up some Perl that perlcritic.com didn't like

scottchiefbaker:
I ran all of `header_clean.pl` through perlcritic.com on level 4 stern (anything less was WAY too picky) and it found one error: a sub without a return value. This sub operates pass-by-reference so it won't affect anything, but it will make Perl critic happy.


paulirish:
nice nice

paulirish:
filed https://github.com/so-fancy/diff-so-fancy/issues/118 in case we want to make this a real thing. :)

PR #117, opened 2016-03-08
Mnemonic prefix fixes

### LINKED ISSUE #103: Files are considered to be renamed with `diff.mnemonicprefix`
% git -c diff.mnemonicprefix=true-c pager.diff=diff-so-fancy diff
    …
    renamed: i/luaa.c to w/luaa.c
    …

That's because it's only looking for `a` and `b` there, but should discard the first path segment altogether probably.

scottchiefbaker:
This should fix #103

### DIFF
--- a/libs/header_clean/header_clean.pl
+++ b/libs/header_clean/header_clean.pl
@@ -29,12 +29,12 @@ for (my $i = 0; $i <= $#input; $i++) {
 	########################################
 	# Find the first file: --- a/README.md #
 	########################################
-	} elsif ($line =~ /^$ansi_sequence_regex--- ([abiwco]\/)?(.+?)(\e|$)/) {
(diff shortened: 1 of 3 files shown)


blueyed:
This PR missed the `$last_file_seen =~ s|a/||; # Remove a/` line above (https://github.com/so-fancy/diff-so-fancy/blob/2f743f07941e744b2f24218dc422f26512823549/lib/diff-so-fancy.pl#L59)

scottchiefbaker:
You're right! It still passes all the tests, so I'm thinking that code may be redundant now. Specifcally we `basename()` it down below, which probably strips off that `a/` anyway? I'm going to leave it as is for now until we know for sure it breaks something.

blueyed:
Exactly.. I've followed up in https://github.com/so-fancy/diff-so-fancy/pull/146.

PR #121, opened 2016-03-09
improve report bugs

stevemao:


scottchiefbaker:
:+1: I love it... Using base64 is a great idea to preserve those escape sequences.

Only note though... `pbcopy` is OS X only, so the script will error out on Linux and Windows.

paulirish:
on pbcopy.. looks like there's been a few solutions. http://superuser.com/questions/288320/whats-like-osxs-pbcopy-for-linux

this ~polyfill script appears to be the most comprehensive [i can find](https://github.com/search?l=bash&p=1&q=pbcopy+xsel+xclip&type=Code&utf8=%E2%9C%93): https://github.com/pekepeke/dotfiles/blob/master/bin/pbcopy-wrapper

paulirish:
most people wont have a clone so we'lll have to curl down this script from rawgit.com

scottchiefbaker:
`pbcopy` just reads in stdin and puts it on the clipboard?

paulirish:
@scottchiefbaker yup

stevemao:
I'm thinking we could also include related `git config` information. We can add an issue template asking user to run the command before creating a ticket.

stevemao:
I have made the changes. WDYT?

PR #123, opened 2016-03-09
Move some functions to perl

scottchiefbaker:
We're down to six calls to **sed** after this patch.

**Note:** All tests still pass.

### DIFF
--- a/libs/header_clean/header_clean.pl
+++ b/libs/header_clean/header_clean.pl
@@ -128,11 +128,10 @@ sub strip_empty_first_line {
 # Remove + or - at the beginning of the lines
 sub strip_leading_indicators {
 	my $array            = shift(); # Array passed in by reference
-	my $line_count       = scalar(@$array);
 	my $ansi_color_regex = qr/(\e\[[0-9]{1,3}(?:;[0-9]{1,3}){0,3}[mK])?/;
 
-	for (my $i = 0; $i < $line_count; $i++) {
-		$array->[$i] =~ s/^(${ansi_color_regex})[+-]/$1 /;
+	foreach my $line (@$array) {
+		$line =~ s/^(${ansi_color_regex})[+-]/$1 /;
 	}
 
 	return 1;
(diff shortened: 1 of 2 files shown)


libs/header_clean/header_clean.pl
```diff
@@ -106,12 +106,11 @@
 
 # Courtesy of github.com/git/git/blob/ab5d01a/git-add--interactive.perl#L798-L805
 sub parse_hunk_header {
-    my ($line) = @_;
-    my ($o_ofs, $o_cnt, $n_ofs, $n_cnt) =
-    	$line =~ /^@@+(?: -(\d+)(?:,(\d+))?)+ \+(\d+)(?:,(\d+))? @@+/;
-    $o_cnt = 1 unless defined $o_cnt;
-    $n_cnt = 1 unless defined $n_cnt;
-    return ($o_ofs, $o_cnt, $n_ofs, $n_cnt);
+	my ($line) = @_;
```
paulirish: oops. sorry about the spaces. :)

scottchiefbaker: I think I'm the ONLY one that uses tabs, so no big deal.

libs/header_clean/header_clean.pl
```diff
@@ -124,3 +124,84 @@ sub strip_empty_first_line {
...
+		$line =~ s/^(${ansi_color_regex})[+-]/$1 /;
+	}
+
+	return 1;
+}
+
+# Remove the first space so everything aligns left
+sub strip_first_column {
+	my $array            = shift(); # Array passed in by reference
+	my $ansi_color_regex = qr/(\e\[[0-9]{1,3}(?:;[0-9]{1,3}){0,3}[mK])?/;
```
paulirish: can we define this variable just once?

scottchiefbaker: We'd have to define it globally and call it that way. Is that what you're thinking? I can make that change if you want.

paulirish: yeah. especially for this particular regex.

paulirish:
I'm curious if we can get away with this work without iterating over all lines within each `sub`. 

I also just generally feel better about things when each modification is not made distinctly. 

aside from this point, i'm feeling good about this work and am happy to move it forward.

@scottchiefbaker can you point out how `get_git_config` will be used? it can land in a separate commit, but i think we just want to understand the general workflow.

scottchiefbaker:
There are two different approached in my Perl now.

1. A big foreach over each line that handles things like removing file perms and header chunks
2. A sub for cleaning the data like removing the + / -

The first approach is probably faster, but the code is much harder to read, follow, and maintain. The second approach is much more readable and digestible but incurs more overhead because we iterate over the whole input several times.

Ideally I'd like to move more towards the second method if possible, but I could be swayed with a good argument. Bear in mind, that right now we're iterating over the entire file ~6 times, one for each sed call. 

Ultimately your question is "are we efficient?" and I think the answer is: we're efficient enough. Is there room for improvement, probably.

paulirish:
Merging this. Let's move the gitconfig stuff into a new issue.

PR #128, opened 2016-03-15
rename header_clean.pl to diff-so-fancy.pl

paulirish:
Since now most of the work is in the perl, it should be a little more first-class.

this is a cosmetic change, only.

@scottchiefbaker work for you?


scottchiefbaker:
I was thinking something along the same lines, you just beat me to it. Works for me. :+1:

paulirish:
cool

we could also move diff-highlight into lib/diff-highlight  but no rush

Issue #126, opened 2016-03-09
Support 3-way merge commits


paulirish:
Currently DSF doesn't correctly format output when viewing 3-way merge commits. There shouldn't be errors, but we misformat the diff, as there can be an extra layer of indentation.

Example: 
<a href="https://cloud.githubusercontent.com/assets/39191/13648024/a3fbad72-e5eb-11e5-8921-7dc211648209.png"><img src="https://cloud.githubusercontent.com/assets/39191/13648024/a3fbad72-e5eb-11e5-8921-7dc211648209.png" height=400></a>

## Impact
* **`git show` on merge commits**
 * example: `git show 74804e377` in the DSF repo.
* **`git diff` across a merge commit (sometimes)**
  * example: ` git diff 91bc769c02  59aebbd379 `  or just `git diff 59aebbd379` 
* **NOT `git diff` on (unstaged) changes**
  * afaik, impossible. So our exposure to this bug is limited.

We discussed this a bit more over in scott's fork: https://github.com/scottchiefbaker/diff-so-fancy/pull/1#issuecomment-194506753 

06b9421b814d80216d7f8a74621aba4b4979a1db provides a test we can use once we're ready to fix this.

<hr>

Currently, we've decided to defer and address this bug later.

lode:
As far as I know this more commonly happens with:

- `git diff` after ending up in a merge conflict
- `git diff` after ending up in a stash pop conflict

stevemao:
Just to summarise the discussion in scott's fork, there might be an n-way merge :) So all we need to do is to count how many number sequences in the header between `@@@`s

Git is really crazy... I wonder how many people actually cares about n-way merge and count the beginning spaces. lol

lode:
But how do you want to solve it? It is not really nice to remove the +/- signs as we do with normal diffs.
Maybe DSF should even revert to removing them at all for n-way merges? So by default they are hidden, and for n-way merges, all are shown as normal.

paulirish:
git doesn't support 4 way merge, so the 3 way is the most complex we'll
have to deal with.

scottchiefbaker:
If there is a three way merge, will the entire diff be a three-way merge? Or is is possible to have ONE file be a three way merge, and another file just be a two-way merge? I'm going to tackle this today, shouldn't be too gnarly.

paulirish:
scott is this good to close now?

PR #132, opened 2016-03-17
Properly remove the right number of leading columns in a three way merge

scottchiefbaker:
Test is also included so we'll know if we break it in the future.

### DIFF
--- a/lib/diff-so-fancy.pl
+++ b/lib/diff-so-fancy.pl
@@ -128,28 +128,6 @@ sub strip_empty_first_line {
 	return 1;
 }
 
-# Remove + or - at the beginning of the lines
-sub strip_leading_indicators {
-	my $array = shift(); # Array passed in by reference
-
-	foreach my $line (@$array) {
-		$line =~ s/^(${ansi_color_regex})[+-]/$1 /;
-	}
-
-	return 1;
-}
-
-# Remove the first space so everything aligns left
-sub strip_first_column {
-	my $array = shift(); # Array passed in by reference
-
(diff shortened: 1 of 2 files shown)


scottchiefbaker:
One thing I changed was I combined `strip_leading_indicators` and `strip_first_column`. I can't think of a use case where we'd want to remove the pluses and minus indicators and not ALSO that first column that's just a space. 

This should also speed it up a bit because it's one less iteration over the file.

paulirish:
Over in #126 i had a few examples:

git show 74804e377
git diff 91bc769c02 59aebbd379

Can you verify these diffs are showing what you'd expect? At first glance, I'm not seeing what I thought we'd show.

scottchiefbaker:
`git show 74804e3` looks perfect

`git diff 91bc769 59aebbd` includes some leading `+` and `-` because those are actually part of the body of the diff. I think this is correct behavior.

`cat test/fixtures/complex-hunks.diff | diff-so-fancy` will give you a good idea of what we're stripping off. 

The *only* caveat to this is that my detection assumes the same indent level (three-way vs two-way merge) for the entire diff. So **if** it's possible that part of the diff is two way and part is three-way this will probably not remove the correct amount. I don't know three-way merges well enough to know if this is a real world scenario or not.

scottchiefbaker:
Any updates on this?

paulirish:
merged! :D

paulirish:
sorry for the delay. @scottchiefbaker in these sort of cases, i'm pretty comfortable with you merging on your own.

PR #131, opened 2016-03-17
Try and be smart about what line the diff starts on

scottchiefbaker:
I've been testing this for a couple of days now and I'm pretty confident it's accurate. This should fix #120. 

```
Change on line # gives this output
1: @@ -1,4 +1,4 @@
2: @@ -1,5 +1,5 @@
3: @@ -1,6 +1,6 @@
4: @@ -1,7 +1,7 @@
5: @@ -2,6 +2,7 @@
6: @@ -3,7 +3,7 @@

Starting from the end (last line (10), second to last line (9))...
-1: @@ -7,4 +7,4 @@
-2: @@ -6,5 +6,5 @@
-3: @@ -5,6 +5,6 @@
-4: @@ -4,7 +4,7 @@
```

PR #134, opened 2016-03-21
add test for #131

stevemao:
Sorry for taking so long...


scottchiefbaker:
This looks good. Thanks.

PR #135, opened 2016-03-22
add a newline after header

stevemao:
Prepare for #35


stevemao:
needs more tests for #35 with the up coming git. Closing for now.

PR #136, opened 2016-03-28
Move "highlight context line" to Perl (one less sed call)

scottchiefbaker:


paulirish:
nice nice. 

well done

PR #139, opened 2016-03-30
update readme to mention the lib dependency

bronson:
Also added an example listing as a separate commit.  It might not be necessary, feel free to omit.


scottchiefbaker:
This is a good clarification. I'm going to merge this.

Issue #129, opened 2016-03-15
Allow runtime changing of diff-so-fancy options by reading from ~/.gitconfig


scottchiefbaker:
Per @paulirish I'm creating a new issue for gitconfig stuff...

-----

I just posted a starter branch on my repo. Specifically the code would look something like:

https://github.com/scottchiefbaker/diff-so-fancy/blob/gitconfig/libs/header_clean/header_clean.pl#L10

and

https://github.com/scottchiefbaker/diff-so-fancy/blob/gitconfig/libs/header_clean/header_clean.pl#L171

It allows us to set run-time options to enable/disable certain behaviors. Still preliminary, but it's already reading changeHunkIndicators, stripLeadingSymbols, and markEmptyLines from your ~/.gitconfig. All these names are up for debate by the way, I just picked something simple.

scottchiefbaker:
This would allow us to solve #69

paulirish:
Nice. 

Is there convention for making sure all options that override default behavior must be set to "true"?  Wording is hard, but we may want to sort that out first.

scottchiefbaker:
I was thinking the opposite. The current toggleable (is that a word) options are:

```
stripLeadingSymbols
markEmptyLines
changeHunkIndicators
```

All default to true, and you explicitly turn OFF the behavior? I can see it either way.

paulirish:
Perhaps they might look something like...

```yaml
[diff-so-fancy]
   leadingSymbols = always
   emptyLineMarkers = never
   adjustHunkDisplay = never
```

And they default to 'auto'

paulirish:
(revised the naming in my comment)

scottchiefbaker:
My thinking was... without ANY config dsf performs as it does now (strips symbols, adjust hunks, etc). You "opted in" when you install dsf, in other words all the options default to "on" (true). If you want to disable one of the runtime options you add that to your config and set it to false.

The always/never syntax only makes sense to me if there is a third option (only in GUI, etc), otherwise we should just use true/false.

paulirish:
k sounds good.

```yaml
[diff-so-fancy]
   stripLeadingSymbols = false
   markEmptyLines = false
   adjustHunkIndicators = false
```

stevemao:
`markEmptyLines` shouldn't be an option I think... Whenever `stripLeadingSymbols` is `true`, `markEmptyLines` must be `true`. Vice versa.

scottchiefbaker:
[...]

PR #141, opened 2016-03-31
This is my first attempt at reading items from ~/.gitconfig

scottchiefbaker:
This should address #69 and #129 

I don't think this is final, but it's ready for testing. We should probably discuss naming, and what section of the gitconfig we want to use. By default all options are enabled (like they were previous to this commit), but they can be turned *off* with the following commands:

```
git config --unset --global diff-so-fancy.markEmptyLines
git config --unset --global diff-so-fancy.changeHunkIndicators
git config --unset --global diff-so-fancy.stripLeadingSymbols
```
    
```
git config --bool --global diff-so-fancy.markEmptyLines false
git config --bool --global diff-so-fancy.changeHunkIndicators false
git config --bool --global diff-so-fancy.stripLeadingSymbols false
```

### DIFF
--- a/lib/diff-so-fancy.pl
+++ b/lib/diff-so-fancy.pl
@@ -7,9 +7,9 @@ use File::Basename;
 my $remove_file_add_header    = 1;
 my $remove_file_delete_header = 1;
 my $clean_permission_changes  = 1;
-my $change_hunk_indicators    = 1;
-my $strip_leading_indicators  = 1;
-my $mark_empty_lines          = 1;
+my $change_hunk_indicators    = git_config_boolean("diff-so-fancy.changeHunkIndicators","true");
+my $strip_leading_indicators  = git_config_boolean("diff-so-fancy.stripLeadingSymbols","true");
+my $mark_empty_lines          = git_config_boolean("diff-so-fancy.markEmptyLines","true");
 
 #################################################################################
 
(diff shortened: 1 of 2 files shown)


scottchiefbaker:
It should be noted that I've been dog fooding this for several weeks now without issue.

paulirish:
Is ANSI reveal yours? If so then it can be in lib
Also thinking that we could split the gitconfig methods into a config.pl
Otherwise this is looking Clean. 

And yah I think we could start bikeshedding the names. :)

scottchiefbaker:
Yes ANSI reveal is mine... I'll probably release it separately later, so I left it in third party so it's more "standalone" and not d-s-f specific.

scottchiefbaker:
This has been working well for me. Any thoughts on if we're ready to merge this?

@stevemao, @paulirish, @blueyed

blueyed:
@scottchiefbaker 
My Perl foo is very low - sorry, but the idea is good!

Just a thought: it would be useful to easily disable d-s-f.
Currently I am using things like `gdraw='git -c pager.diff="$(git config core.pager)" diff'` to bypass it.
Bypassing will be always useful in case you want to copy'n'paste some diff hunk.

stevemao:
@blueyed https://github.com/so-fancy/diff-so-fancy#opting-out

stevemao:
@scottchiefbaker I don't really get [this comment](https://github.com/so-fancy/diff-so-fancy/issues/129#issuecomment-197455815) yet... but if it works well feel free to merge this.

blueyed:
@stevemao 

> git --no-pager diff

Thanks!
Although that cannot be applied/used at the end of an existing alias..  so I cannot replace `gd` with `gd --no-pager` when needed, but need to use `git --no-pager diff` instead.  A `gnp` alias might make sense, but still it's hard to transform from the original command..  suggestions?

stevemao:
@blueyed I don't think we should add a flag in this module... Rather, rely on git commands or shell.

lavoiesl:
Tested locally, works perfectly.

scottchiefbaker:
I've been testing this for about three weeks and it's working great. I'm going to merge this, so we can move on to bigger and better things.

paulirish:
lookin sharp!

matz3:
What are the plans for a new release which includes this feature? Any open points/issues?

Issue #140, opened 2016-03-30
diff-so-fancy slurps entire git log -p output?  very slow


bronson:
Using Feb 7th's diff-so-fancy, `git log -p` and `git log --stat` would display immediately.

Using today's diff-so-fancy, `git log --stat` takes many seconds to complete and `git log -p` hangs for a few minutes and then bombs out.

A guess: it looks like some change in the past 1.5 months made it so diff-so-fancy is now trying to slurp the entire output before displaying the results.

My repo has 4200 commits, so tiny.  This issue appears reproducible in any moderately-sized project.  For example, `git log --stat` on rails/rails shows a blank screen for 10 seconds before showing the log.  Revert to Feb 7th and it displays immediately.

stevemao:
I don't think you should use `diff-so-fancy` for log pager.

As per readme:
```
git config --global pager.diff "diff-so-fancy | less --tabs=4 -RFX"
git config --global pager.show "diff-so-fancy | less --tabs=4 -RFX"
```

scottchiefbaker:
I agree with @stevemao. d-s-f is designed to process small diffs (commits) in to human readable form, not to parse hundreds of thousands of lines. 

My guess for the slowness is that we remove all the leading `+` and `-` from the diff before we process any actual content. It'd be pretty hard to work around that for this one use case.

bronson:
Ah, the recommended config changed on Feb 12: 5f7422aa38c0bd16640f85dc2f02b72f85e8af88

Updated my config to match and all is fine.  Thanks.

bronson:
That said....   it's interesting that Feb 7th's d-s-f could handle being the core.pager and today's can't.

blueyed:
I agree that it would be better if it still could be used with `log`.

It would be helpful to find the commit that caused this (using git-bisect) and then improve upon that.
Even if the removal/processing of all +/- would cause this, this could still be done in a pipeline then probably?!

scottchiefbaker:
I think @bronson found it at a6b5801. That's the commit where we switched most of the backend from sed to Perl so we could do fancier things with the output display. We can investigate the slowness, but I can't make any guarantees as this use case is a little outside the scope of the project.

blueyed:
It should still be possible to not read the whole input, but process it in chunks (e.g. 8kb).

scottchiefbaker:
If I use a 60MB `git log -p` output I see the following:

* 1 second to load the whole input in to memory
[...]

PR #142, opened 2016-04-01
Streamline the input by reading one line at a time

scottchiefbaker:
Before we'd load the entire input in to memory, but this should:

1) Use less memory
2) Output things to the screen a lot faster


scottchiefbaker:
This should address #140 and dramatically speed up processing of large files, and use considerably less memory.

paulirish:
I'm not seeing any issues here. Everything looks good and tbh it doesnt look like we regress code health.

Nice work, @scottchiefbaker !! :+1:

PR #143, opened 2016-04-04
Do not sneak in another dash/plus in format_diff_header

blueyed:
And especially do not match with `*` for this, which might trigger false
positives.

This moves the removal of leading +/- down to where those lines are
printed, instead of applying it globally to the input.

Ref: https://github.com/so-fancy/diff-so-fancy/commit/2f743f07941e744b2f24218dc422f26512823549#commitcomment-16952151

No new tests, but maybe I can get one for 10 innernet points from @scottchiefbaker? :) (https://github.com/so-fancy/diff-so-fancy/commit/2f743f07941e744b2f24218dc422f26512823549#commitcomment-16953657)

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -35,8 +35,8 @@ git_index_line_pattern="^($color_code_regex)index .*"
 format_diff_header () {
   # simplify the unified patch diff header
     $SED -E "/$git_index_line_pattern/{N;s/$git_index_line_pattern\n//;}" \
-    | $SED -E "s/^($color_code_regex)\-\-\-(.*)$/\1$(print_horizontal_rule)\\${NL}\1\-\-\-\-\5/g" \
-    | $SED -E "s/^($color_code_regex)\+\+\+(.*)$/\1\+\+\+\+\5\\${NL}\1$(print_horizontal_rule)/g"
+    | $SED -E "s/^($color_code_regex)\-\-\-(.*)$/\1$(print_horizontal_rule)\\${NL}\1---\5/g" \
+    | $SED -E "s/^($color_code_regex)\+\+\+(.*)$/\1+++\5\\${NL}\1$(print_horizontal_rule)/g"
 }
 
 print_horizontal_rule () {
--- a/lib/diff-so-fancy.pl
+++ b/lib/diff-so-fancy.pl
@@ -35,11 +35,6 @@ while (my $line = <>) {
 		next;
 	}
 
-	# Mark empty line with a red/green box indicating addition/removal
-	if ($mark_empty_lines) {
-		$line = mark_empty_line($line);
-	}
-
 	######################
 	# End pre-processing
 	######################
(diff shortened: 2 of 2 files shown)


scottchiefbaker:
Before we land this let's wait and see what @pauley-unsaturated says. This would revert part of his code, and un-fix whatever he was fixing. Also nothing is "broken" we just work around it in a different way so it's not really a big deal either way.

lib/diff-so-fancy.pl
```diff
@@ -135,6 +130,10 @@
 	# Just a regular line, print it out #
 	#####################################
 	} else {
+		# Remove the correct number of leading " " or "+" or "-"
+		if ($strip_leading_indicators) {
+			$line = strip_leading_indicators($line,$columns_to_remove);
+		}
```
scottchiefbaker: If we're going to move the pre-processing logic for `strip_leading_indicators()` down here we should move `mark_empty_lines()` as well.

pauley-unsaturated:
I'm fine with this.  My original change was to address the fact that we were removing one of the three '+' signs that git diff inserted.  The original matching expression didn't appear to be removing one of them on purpose.  Two '+' or '-' is just fine with me.

diff-so-fancy
```diff
@@ -35,8 +35,8 @@ git_index_line_pattern="^($color_code_regex)index .*"
 format_diff_header () {
   # simplify the unified patch diff header
     $SED -E "/$git_index_line_pattern/{N;s/$git_index_line_pattern\n//;}" \
-    | $SED -E "s/^($color_code_regex)\-\-\-(.*)$/\1$(print_horizontal_rule)\\${NL}\1\-\-\-\-\5/g" \
-    | $SED -E "s/^($color_code_regex)\+\+\+(.*)$/\1\+\+\+\+\5\\${NL}\1$(print_horizontal_rule)/g"
+    | $SED -E "s/^($color_code_regex)\-\-\-(.*)$/\1$(print_horizontal_rule)\\${NL}\1---\5/g" \
+    | $SED -E "s/^($color_code_regex)\+\+\+(.*)$/\1+++\5\\${NL}\1$(print_horizontal_rule)/g"
```
pauley-unsaturated: However: does this work? + is a special character in SED.  I guess since this is in the second pattern, kleene plus and kleene star don't apply so they're treated as normal characters?

scottchiefbaker: I think as the second parameter there fine because you're not talking about patterns.

blueyed: Yes.

lib/diff-so-fancy.pl
blueyed: Not so easy.. because it also strips escape sequences, and that's expected above already.
I'd say to keep it there for now.

scottchiefbaker: Then my vote would be to keep them together for readability. Ultimately `strip_leading_indicators()` is getting called on EVERY output line anyway. Moving it isn't really saving us much in terms of overhead.

blueyed: The problem with `strip_leading_indicators` is that it will also strip one level from the `---/+++` markers.  That was the main reason for moving it.

blueyed: Okay, I was wrong: it can be moved down.. done in a separate commit.

stevemao:
A test is necessary. Just add what was broken before this change. @blueyed

blueyed:
@stevemao 
I think it's covered with the test in https://github.com/so-fancy/diff-so-fancy/commit/c3a141b7e60c597a38426fa709b6a0e3bd7fe766.

PR #146, opened 2016-04-04
Handle mnemonicprefix with last_file_seen

blueyed:
This is not really required currently, because `basename` is used on it,
but it is good to have it fixed nonetheless.


scottchiefbaker:
We already have a test for this "mnemmonic prefix", and we pass that. That means we're parsing the header correctly already right? Am I missing something here?

blueyed:
> Am I missing something here?

The special case is for when a file is deleted: then `$last_file_seen` is not updated from the `+++` line (https://github.com/so-fancy/diff-so-fancy/blob/2f743f07941e744b2f24218dc422f26512823549/lib/diff-so-fancy.pl#L73-L75), and the one patched in this PR would be used.

There was no test for that as far as I can see, and the test would pass also currently (without the patch), because `basename` strips off the `c/`.

PR #147, opened 2016-04-04
Integration of #143, #144, #146

blueyed:
Since they conflict I've thought that this might be helpful, in case they all are accepted/make sense.

### DIFF
--- a/lib/diff-so-fancy.pl
+++ b/lib/diff-so-fancy.pl
@@ -35,11 +35,6 @@ while (my $line = <>) {
 		next;
 	}
 
-	# Mark empty line with a red/green box indicating addition/removal
-	if ($mark_empty_lines) {
-		$line = mark_empty_line($line);
-	}
-
 	######################
 	# End pre-processing
 	######################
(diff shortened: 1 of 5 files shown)


lib/diff-so-fancy.pl
```diff
@@ -56,16 +47,17 @@
 	#########################
 	if ($line =~ /^${ansi_color_regex}diff --(git|cc) (.*?)(\s|\e|$)/) {
 		$last_file_seen = $5;
-		$last_file_seen =~ s|a/||; # Remove a/
+		$last_file_seen =~ s|\w/||; # Remove a/ (and handle diff.mnemonicPrefix).
```
scottchiefbaker: I think this should be change to:

    $last_file_seen =~ s|^\w/||; # Remove leading "a/" from the found file

Just add an ^ anchor

scottchiefbaker:
This looks good to me. I made one small suggestion. If you're ok with that change (and all the tests pass) I'm fine with merging this.

blueyed:
Great.
Added two fixups (still green), and then rebased it - hopefully it will still be green.

scottchiefbaker:
This looks great to me. If the tests all pass I'll merge this in. I think these are good changes and will result in a more accurate d-s-f.

scottchiefbaker:
@stevemao, @paulirish any objections before I merge this?

stevemao:
@scottchiefbaker I'm not very update-to-date with this repo so feel free to merge it :smile:

PR #153, opened 2016-05-12
Fix Use of uninitialized $value in substitution (s///)

jhermsmeier:
Got the following error since `0.9.0`:
```
$ git diff
Use of uninitialized value $value in substitution (s///)
at /usr/local/lib/node_modules/diff-so-fancy/lib/diff-so-fancy.pl line 216.
```
This attempts to fix it, by only assigning `$raw->{$key}` if `$value` has a value.

### DIFF
--- a/lib/diff-so-fancy.pl
+++ b/lib/diff-so-fancy.pl
@@ -212,10 +212,11 @@ sub git_config {
 
 	my $raw = {};
 	foreach my $line (@$out) {
-		my ($key,$value) = split("=",$line,2);
-		$value =~ s/\s+$//;
-
-		$raw->{$key} = $value;
+		if ($line =~ /=/) {
+			my ($key,$value) = split("=",$line,2);
+			$value =~ s/\s+$//;
+			$raw->{$key} = $value;
+		}
 	}
 
 	# If we're given a search key return that, else return the hash


scottchiefbaker:
This fix seems logical to me. I'm a little more concerned about what is actually triggering this condition. Can include a raw diff output that I can run through d-s-f to recreate this error?

jhermsmeier:
If I understood the code around this correctly, this is not related to the diff going through it, as it looks up the config options in the user's `.gitconfig` – and fails because of the empty `$value` when the user has no (or not all) dsf related options set in the `.gitconfig`.

scottchiefbaker:
Ahhh good catch. I misread the snippet. Looking at that section of code it should only fail if your `git config --list` output does not contain `key=value` type of data. Can you post the output of your `git config --list` (you can sanitize any sensitive info) so I can verify my thinking?

jhermsmeier:
Sure, I put the output of `git config --list` into a gist, for brevity: https://gist.github.com/jhermsmeier/33d1402a591876acaa2340f0f6d976ff
There are two lines that do not have a `key=value` format, due to `\n`s in two aliases – guess I have a malformatted config then?

scottchiefbaker:
Is that output truncated, or are lines 63 and 136 broken out like that?

jhermsmeier:
They are output like that – surprised me as well. I just fixed my newline escapes in my aliases – which fixes this issue as well. Guess it'd still be a good idea to check for this, to not render dsf unusable if people have slightly malformatted `.gitconfig`s (like I did)?

scottchiefbaker:
Ya I think the prudent thing to do is to check if the line has a `=` in it before we do the split. Just wrap that whole split section in:

```
if ($line =~ /=/) {
    # Do the split stuff
}
```

If you wanna modify your pull request to something like I'll review it and then land the fix.

jhermsmeier:
Yup, updated it to check for `$line =~ /=/` now

scottchiefbaker:
Looks good to me. I'll merge this.

jhermsmeier:
Sweet, thanks!

PR #154, opened 2016-05-13
Suggest configuring for git log as well

alex:


paulirish:
my logs dont have diffs within them. do you have them configured as such?

alex:
@paulirish `git log -p` :-)

paulirish:
oh word, that's right..  :)  _(btw hi hi. big fan of all the work you've up to.)_

----------------------

We used to have log included in the config by default, but #42 is what turned the tide.
I think the challenge here is performance and matching user expectations. 

.. Okay I say performance but I just tested things...

```sh
git log -p  | wc -l
#  429169
 git log -p  | diff-so-fancy | less --tabs=1,5 -RFX
# displayed in <130ms
```

That's crazy fast. Same fast performance in the Chromium repo. dayum. (Well done @scottchiefbaker!)

Well in that case, it'd be totally open to adding this back to the recommended configuration. 

Trying it out a little, it seems like the parser has no particular issues with the log. Feels pretty safe.

paulirish:
@scottchiefbaker @stevemao thoughts?

@jamischarles can you explain what lead to #42? do you have a separate pager for log configured?

scottchiefbaker:
If performance is good (it should be), then I don't have a problem with this.

stevemao:
> If performance is good (it should be), then I don't have a problem with this.

I suggest to revert #44

paulirish:
SG.

Yeah I think we've come full circle on the config. so @alex it'd be best to adapt your PR to be close to a revert of https://github.com/so-fancy/diff-so-fancy/pull/44/files

jamischarles:
@paulirish If I recall correctly (it was a while ago) things like `git show` didn't work as expected (just blank I think). Here's my confit for reference: https://github.com/jamischarles/dotfiles/blob/master/gitconfig.sym

paulirish:
@jamischarles gotcha. The project's been rewritten almost completely since then so I wouldn't expect you'd have these challenges again. But we can keep an eye on it.

alex:
@paulirish Ok, I think I got this right.

stevemao:
LGTM

Issue #155, opened 2016-05-16
Highlight binary files appropriately


alex:
Normal diff sections are nicely formatted like:

```
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
modified: administration/project_summary.txt
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
```

with bright white text. However binary files have a far more plain, and easily missed:

```
Binary files a/administration/priorities.pdf and b/administration/priorities.pdf differ
```

in boring gray.

It'd be awesome if that `Binary files [a] [b] differ` line was nicer looking and better distinguished from the rest of the text

scottchiefbaker:
Good catch... I think I was able to fix that, while adding some performance improvements as well. Try replacing your `lib/diff-so-fancy.pl` with this one: 

https://github.com/scottchiefbaker/diff-so-fancy/blob/binary-file-header/lib/diff-so-fancy.pl

If this does what you want it to I'll send a PR to get it landed in the next version.

scottchiefbaker:
Sorry I got confused... there is another script that I updated also. Probably easiest to check out the whole repo and see if it fixes the problem or not.

https://github.com/scottchiefbaker/diff-so-fancy/tree/binary-file-header

scottchiefbaker:
@alex have you had a chance to test this yet?

alex:
Sorry, no I haven't.

williamboman:
Looks good.

<img width="1550" alt="screenshot 2016-05-20 at 13 58 58" src="https://cloud.githubusercontent.com/assets/6705160/15427144/44720c94-1e93-11e6-82ab-4a2ae1f71aba.png">
<img width="1550" alt="screenshot 2016-05-20 at 14 00 17" src="https://cloud.githubusercontent.com/assets/6705160/15427145/44727422-1e93-11e6-98c0-1a31a9946076.png">

scottchiefbaker:
Great. I'll get the PR ready.

stevemao:
fixed by #156

alex:
awesome, thanks.

PR #156, opened 2016-05-20
Properly display binary file modifications

scottchiefbaker:
This also moves `horizontal_rule()` to Perl. Three less calls to sed so it should be quicker too.


scottchiefbaker:
Forgot to mention, this should fully address #155

paulirish:
what a beauty of a PR. lookin sharp!

PR #158, opened 2016-05-24
don't print characters that we know will be incompatible with less(1)

williamboman:
This is very corner-case and maybe not that interesting. (see https://github.com/so-fancy/diff-so-fancy/issues/72#issuecomment-219591245)

screenshot 2016-05-24 at 19 53 48

<sup>(the dashes do not overflow like in the screenshot, created a new pane after)</sup>

### DIFF
--- a/lib/diff-so-fancy.pl
+++ b/lib/diff-so-fancy.pl
@@ -275,6 +275,22 @@ sub in_unit_test {
 	}
 }
 
+sub get_less_charset {
+	my @less_char_vars = ("LESSCHARSET", "LESSCHARDEF", "LC_ALL", "LC_CTYPE", "LANG");
+	foreach (@less_char_vars) {
+		return $ENV{$_} if defined $ENV{$_};
+	}
+}
+
+sub should_print_unicode {
+	if (-t STDOUT) {
+		# Always print unicode chars if we're not piping stuff, e.g. to less(1)
+		return 1;
+	}
+	# Otherwise, assume we're piping to less(1)
+	return get_less_charset() =~ /utf-?8/i;
+}
+
 # Return git config as a hash
 sub get_git_config_hash {
 	my $out = git_config_raw();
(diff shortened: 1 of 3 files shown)


lib/diff-so-fancy.pl
```diff
@@ -376,7 +379,7 @@ sub horizontal_rule {
 	# em-dash http://www.fileformat.info/info/unicode/char/2014/index.htm
 	#my $dash = "\x{2014}";
 	# BOX DRAWINGS LIGHT HORIZONTAL http://www.fileformat.info/info/unicode/char/2500/index.htm
-	my $dash = "\x{2500}";
+	my $dash = $should_print_unicode_chars ? "\x{2500}" : "-";
```
scottchiefbaker: For consistency (I don't use tertiaries) can you change this to an if/then statement.

lib/diff-so-fancy.pl
```diff
@@ -14,6 +14,9 @@
 my $strip_leading_indicators  = git_config_boolean("diff-so-fancy.stripLeadingSymbols","true");
 my $mark_empty_lines          = git_config_boolean("diff-so-fancy.markEmptyLines","true");
 
+# see https://git.io/vrOF4
```
scottchiefbaker: Can you add a comment here explaining what we're doing. Something like

    # Check to see if we have appropriate env variables to do unicode

scottchiefbaker:
This is a decent addition to avoid weird bugs in the future. If you fix a couple of the things I mentioned I'll merge this also.

Also +10 bonus points for writing tests to check this

williamboman:
Updated with a much cleaner solution

scottchiefbaker:
EXCELLENT!

I **always** defer to longer and easier to read code than short complex code. This is a perfect solution in my mind. Thanks for the excellent fix.

PR #161, opened 2016-05-25
test/diff-highlight: avoid word splitting

williamboman:
Looked a bit scary as-is

### DIFF
--- a/test/diff-highlight.bats
+++ b/test/diff-highlight.bats
@@ -9,9 +9,9 @@ load 'test_helper/util'
 output=$( load_fixture "ls-function" | $diff_so_fancy )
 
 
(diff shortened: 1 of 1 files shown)


scottchiefbaker:
Just some basic clean up stuff? @stevemao this looks harmless to me, what do you think?

stevemao:
What linter are you using @williamboman ?
@paulirish I roughly remember we were going to use a linter for bats?

williamboman:
> What linter are you using @williamboman ?

None.

scottchiefbaker:
This is harmless, and it's been open for a while. I'm going to merge this, you can flog me later if you don't like it.

PR #163, opened 2016-05-25
Fix up 'dsf' git alias example in readme

Nephyrin:
- Fix quoting around GIT_PREFIX
- Use && to fail early if the cd does
- Use `git config --global alias.dsf`, which git recommends over
  hand-editing .gitconfig (and makes it easy to adapt the command to
  per-repository configs by dropping --global)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -24,9 +24,9 @@ git diff --color | diff-so-fancy
 git config --global core.pager "diff-so-fancy | less --tabs=4 -RFX"
 ```
 
-However, if you'd prefer to do the fanciness on-demand with `git dsf`, drop an alias in your `~/.gitconfig`:
+However, if you'd prefer to do the fanciness on-demand with `git dsf`, add an alias to your git config:
 ```shell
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -24,9 +24,9 @@ git diff --color | diff-so-fancy
 git config --global core.pager "diff-so-fancy | less --tabs=4 -RFX"
 ```
 
-However, if you'd prefer to do the fanciness on-demand with `git dsf`, drop an alias in your `~/.gitconfig`:
+However, if you'd prefer to do the fanciness on-demand with `git dsf`, add an alias to your git config:
```
scottchiefbaker: I like the change to "add" but I prefer keeping the `~/.gitconfig` part.

scottchiefbaker:
I'm 99% fine with this... anyone writing/fixing the documentation for us saves me time. I had a minor suggestion about the wording for "git config", which I could use a second opinion on. If we can come to a consensus on that I'll be happy to merge this.

OJFord:
I agree with referencing `~/.gitconfig` - it points anyone who doesn't know anything about configuring git in the right direction, and anybody with more experience will know full well that there exist other ways and places for them to add the configuration; it won't confuse them.

PR #167, opened 2016-06-14
issues35 update readme with interactive.diffFilter

tomviner:
Note in the readme about the resolution of https://github.com/so-fancy/diff-so-fancy/issues/35

In git 2.9 you can now use diff-so-fancy in `git add --patch`.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -29,6 +29,11 @@ However, if you'd prefer to do the fanciness on-demand with `git dsf`, drop an a
 dsf = "!f() { [ \"$GIT_PREFIX\" != \"\" ] && cd "$GIT_PREFIX"; git diff --color $@ | diff-so-fancy | less --tabs=4 -RFX; }; f"
 ```
 
+As of **git 2.9** you can also fancy up the output of interactive patch selection (`git add/checkout/stash/reset --patch`). This requires the `--patch-mode` option to ensure everything lines up. Apply it like this:
+```shell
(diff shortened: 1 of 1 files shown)


scottchiefbaker:
This looks good to me. I'm fine with merging this if we can get a second opinion.

tomviner:
This PR should be put on hold until the output isn't garbled see https://github.com/so-fancy/diff-so-fancy/issues/35#issuecomment-225954170 for explanation.

OJFord:
LGTM if update with a note on config for `--patch-mode`. Maybe change "*in* 2.9" to 'from', or 'as of' or something too, but that's pretty pedantic of me. 😀

tomviner:
@OJFord thanks for comments. Updated patch. Re-review please.

scottchiefbaker:
@tomviner this looks good to me. I like the wording, it's mergeable now. I think we should wait for #35 to close before we make this change though. What do you think?

I have 95% of a patch to implement #35, we just need a little more testing.

tomviner:
@scottchiefbaker I think we should wait for https://github.com/so-fancy/diff-so-fancy/issues/35 to land, otherwise more people will get the errors.

tomviner:
@scottchiefbaker I don't think the ` | less --tabs=4 -RFX` is needed for `interactive.diffFilter`. Unless you disagree, I'll remove that from the advice.

agauniyal:
status anyone?

scottchiefbaker:
@agauniyal I think this is a CANTFIX because of the line rearranging that `git --patch` does. There is more information in #35.

Issue #165, opened 2016-06-09
encoding problems diffing latin1 text


adagios:
Hey, I'm getting `utf8 "\xE9" does not map to Unicode at /usr/local/Cellar/diff-so-fancy/0.9.3/libexec/diff-so-fancy.pl line 33, <> line 7.` when trying to diff a file with a single latin1 `é` caracter.

To reproduce:

```
mkdir test
cd test
git init
vim test.txt # add a single é character and save with :w!++enc=latin1
git add test.txt
git diff
```

installed via homebrew (stable 0.9.3)

scottchiefbaker:
I can confirm this is an issue. I'm not really sure how to fix it though. The problem is easily solved by saving the file as UTF8. I'm 99% sure the problem is that we assume all input is utf8, and this isn't. 

We'd have to detect the encoding of the input (STDIN), but you can't really do that unless you open it first, and we open it in UTF8 mode.

https://github.com/so-fancy/diff-so-fancy/blob/master/lib/diff-so-fancy.pl#L8

I'm open to suggestions, but this seems like a pretty obscure corner case, and may not end up getting solved cleanly.

scottchiefbaker:
The more I think about it... this may be a **CANTFIX**. If your diff output incldues `utf8` *and* `latin1` output (from two different files) there is no way for d-s-f to know that. All d-s-f gets is raw `git diff` output, which every file's diff concatenated together. I hate to say it, but it's 2016 so we may have to default `utf8` as it's the only sane default in a world of varied encodings.

Alternately we could allow a user a one-off to *force* the encoding/decoding (not assume utf8) but that only works if all your input is the **same** encoding.

paulirish:
+1 on taking a utf-8-only stance here.

stevemao:
Can we warn user in such case? Or adding an option (flag)?

adagios:
I'm perfectly fine with assuming UTF8. But I don't think it's acceptable to fail when finding an unexpected character.

Don't detect another encoding, that's difficult and impossible to do with absolute certainty. But don't error out, substitute the failing character with a placeholder one.

I was converting a single file between encodings, and the diff aborted. The regular git diff works as expected, showing garbled chars on the non-UTF8 side.

adagios:
On the other hand, there are perfectly valid reasons to use a non-UTF8 encoding, and you should consider supporting them. Java properties files must be iso-8859-1 for example.

scottchiefbaker:
[...]

PR #168, opened 2016-06-14
Properly handle latin1 encoded text

scottchiefbaker:
We assume input is UTF8 (we have to), so when input is latin1 it causes Perl to barf on an unknown char mapping. This disables warnings so the chars get imported as their hex codes like <E9>. It's not a perfect solution, but it will get us by.


scottchiefbaker:
This should address #165

scottchiefbaker:
This is a pretty benign addition. Just looking for a second opinion before this lands.

PR #169, opened 2016-06-14
Handle all binary file modifications similiar to text files

siiptuo:
Most binary file modifications are displayed correctly thanks to #156. However it only fixed one type of operation. This pull request extends the behaviour to handle add, delete and rename operations.

### DIFF
--- a/lib/diff-so-fancy.pl
+++ b/lib/diff-so-fancy.pl
@@ -78,22 +78,7 @@ while (my $line = <>) {
 			$last_file_seen = $file_2;
 		}
 
-		# If they're the same it's a modify
-		if ($file_1 eq $file_2) {
-			print "modified: $file_1\n";
-		# If the first is /dev/null it's a new file
-		} elsif ($file_1 eq "/dev/null") {
-			print "added: $file_2\n";
-		# If the second is /dev/null it's a deletion
-		} elsif ($file_2 eq "/dev/null") {
-			print "deleted: $file_1\n";
-		# If the files aren't the same it's a rename
-		} elsif ($file_1 ne $file_2) {
-			print "renamed: $file_1 to $file_2\n";
-		# Something we haven't thought of yet
-		} else {
-			print "$file_1 -> $file_2\n";
-		}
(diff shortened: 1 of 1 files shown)


lib/diff-so-fancy.pl
```diff
@@ -78,22 +78,7 @@
...
-		} elsif ($file_2 eq "/dev/null") {
-			print "deleted: $file_1\n";
-		# If the files aren't the same it's a rename
-		} elsif ($file_1 ne $file_2) {
-			print "renamed: $file_1 to $file_2\n";
-		# Something we haven't thought of yet
-		} else {
-			print "$file_1 -> $file_2\n";
-		}
+		print "${\file_change_string($file_1,$file_2)}\n";
```
scottchiefbaker: I'm not a huge fan of the readability of this line. Can you change this to:

    print file_change_string($file_1,$file_2) . "\n";

lib/diff-so-fancy.pl
```diff
@@ -126,11 +111,11 @@
...
-	######################################
-	# Look for binary file changes
-	######################################
-	} elsif ($line =~ /Binary files \w\/(.+?) and \w\/(.+?) differ/) {
-		print "${horizontal_color}modified: $2 (binary)\n";
+	################################
+	# Look for binary file changes #
+	################################
+	} elsif ($line =~ /^Binary files (\w\/)?(.+?) and (\w\/)?(.+?) differ/) {
+		print "${horizontal_color}${\file_change_string($2,$4)} (binary)\n";
```
scottchiefbaker: Same as above... this is needlessly unreadable. Please assign the `file_change_string()` to a variable and then print that out on the next line. Function calls inside of print strings are hard to parse. 

While we're at it, since we're removing the hardcoded "modified:" we can remove the wrapping `{ }` around horizontal color.

I **always** favor readability over terseness.

scottchiefbaker:
I really like this! This makes a lot of sense to me.

I made a couple of minor code formatting suggestions, which if you fix I'll gladly merge this.

scottchiefbaker:
This is clean, and it passes all of our tests. Good work!

Issue #118, opened 2016-03-08
perl critic in CI


paulirish:
we have shellcheck running against our bash, which is great.

I think we should also make sure perlcritic exits without error in our tests. probably can help catch issues.

I've found a few projects that do this.
* [ack2/.travis.yml](https://github.com/petdance/ack2/blob/ec4909776b21ac22546275480770a1cbc800c4ec/.travis.yml) installs from cpan without tests (seems good). the makefile is pretty wild though.
* [binary-com/perl-VolSurface-Utils](https://github.com/binary-com/perl-VolSurface-Utils/commit/5e42217f40df9ee5c8a7f86d54c7eef694d4cae1) added things in one go.

we're in no rush for this, but it'll be cool moving forward.

scottchiefbaker:
:+1:

maxnordlund:
I think this is done.

PR #170, opened 2016-06-14
Add perlcritic to Travis

maxnordlund:
This is a first stab at #118. Sorry for the noise, but it's hard to debug this without running it on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,15 @@
-language: generic
+os: linux
+language: perl
+perl:
+  - 5.18
+  - 5.14
 
 matrix:
   include:
     - os: osx
       osx_image: xcode7.3
-    - os: linux
-      language: perl
-      perl: 5.18
+      language: generic
+      perl: 5.24
 
 before_install:
   - git clone --depth 1 https://github.com/sstephenson/bats.git
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,19 @@
-language: perl
+language: generic
+
+matrix:
+  include:
+    - os: osx
+      osx_image: xcode7.3
+    - os: linux
+      language: perl
+      perl: 5.18
+
 before_install:
   - git clone --depth 1 https://github.com/sstephenson/bats.git
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
       brew update;
-      brew install shellcheck;
+      brew install shellcheck perl cpanminus;
+      mkdir -p ~/perl5;
     fi
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
       sudo add-apt-repository "deb http://archive.ubuntu.com/ubuntu/ trusty-backports restricted main universe";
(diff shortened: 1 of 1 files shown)


maxnordlund:
OK, nice it's running on Travis and fails as it's supposed to. I hope this helps :)

scottchiefbaker:
Perl critic can be VERY critical. When I first implemented Perl critic stuff I had to turn the level down so it wasn't complaining about everything. Perl critic is good, but it's OVERLY cautious, and not realistic. I think we had to set it to 4 (stern) for it to be worthwhile when I did it at:

http://perlcritic.com/

Can we modify this test to use a more realistic level?

maxnordlund:
Yeah, you can add a config, but since I don't know what you need I figured you tell me.

scottchiefbaker:
Looking at the errors Perl Critic found makes me cringe a little:

```    
I/O layer ":utf8" used at line 9, column 1.  Use ":encoding(UTF-8)" to get strict validation.  (Severity: 5)
```
OK this is fixable.

```
Use IO::Interactive::is_interactive() instead of -t at line 272, column 6.  See page 218 of PBP.  (Severity: 5)
```
I'm not going to require a third party module JUST for this one silly check. That would make installation a lot more complex.

```
Warnings disabled at line 9, column 1. See page 431 of PBP. 
```
This is by design, and part of #168, so we're not going to fix this

```
Subroutine "get_less_charset" does not end with "return" at line 267, column 1. See page 197 of PBP. 
```
Ok this one I will fix... it's trivial

The last three are all:
```
Always unpack @_ first at line...
```
which I don't understand, because that's exactly what we're doing. I'm guessing it just doesn't like our formatting?

scottchiefbaker:
I discussed this with the experts on #perl and the consensus is that the default Perl critic rules do not reflect real world, modern Perl. The groups things that using this theme for Perl critic is much more useful and sane: https://metacpan.org/pod/Perl::Critic::Freenode

Can you investigate if that's something we can enable and use on TravisCI? When I run it locally we look good:

```
bakers@yellow(~/github/diff-so-fancy)
:perlcritic --theme freenode lib/diff-so-fancy.pl
lib/diff-so-fancy.pl source OK
```

scottchiefbaker:
I spent the last hour trying to get TravisCI to work with cpanm on OSX and it's not happening. Here is my updated config:

https://github.com/scottchiefbaker/diff-so-fancy/blob/travis/.travis.yml

It does work and run Perl Critic using the more sane them on Linux just fine. I may just set the PC tests to only run on Linux. Thoughts?

maxnordlund:
I've added you as an collaborator to make this easier. Yeah, I'll take a look right away. I guess running perlcritic just on Linux isn't so bad since it should just report the same errors on all platforms. This also means it can run on just Travis.

scottchiefbaker:
Perl Critic itself SHOULD run on OSX just fine. Travis has their OSX locked down pretty well, so it doesn't easily install modules. It may not be worth the time/energy.

Don't let me discourage you if you want to try to get it working though.

maxnordlund:
After fighting with inline if statements for a while, I realized that the OSX image was way old. So by upgrading it, we upgrade perl and all is good in wonderland. The brew installed perl didn't seem to be picked up by Travis, so I'll try without it.

maxnordlund:
Scratch that, it didn't work so I'll revert to installing perl again. Maybe it's just in before install it didn't seem to work? Anyway, this should be fine since it's how https://travis-ci.org/so-fancy/diff-so-fancy/jobs/137891671 was setup.

maxnordlund:
OK, it's finally behaving. There were a few tricks needed, as written in the commit message. First use a newer version of OSX, then manually install _both_ `perl` and `cpanm`, and lastly for whatever reason, `mkdir -p ~/perl5` is needed or it will explode.

@scottchiefbaker I think this is good to go now.

scottchiefbaker:
I see it's set as generic now, instead of Perl? Will the affect our ability to test on different versions of Perl? I'm thinking older versions may be more inclined to complain about things, so it would be nice to test against older Perls.

maxnordlund:
You can add more in the build matrix, and maybe you can tweak it further down the line, but if I didn't do this it started two 5.18 Linux builds.

maxnordlund:
I can try one thing though, which might work, but I had a lot of trouble making OSX behave as you I'm sure you know.

maxnordlund:
This seems to make the build matrix better, let us see if OSX wants to play ball.

maxnordlund:
Ah yeah, it's a go time for this PR, don't you think so?

scottchiefbaker:
I see... so we can test against diff versions of Perl on Linux, but on OSX it's the installed system one. I'm OK with that, OSX is a pain. If you can add Perl 5.24, 5.22, and 5.20 to the list I'll land this.

maxnordlund:
Done.

PR #171, opened 2016-06-14
Do not use hardcoded "/usr/bin/perl" path

myfreeweb:
Not all systems install Perl into `/usr/bin`!

### DIFF
--- a/lib/diff-so-fancy.pl
+++ b/lib/diff-so-fancy.pl
@@ -1,4 +1,4 @@
-#!/usr/bin/perl
+#!/usr/bin/env perl
 
 use strict;
 use warnings FATAL => 'all';


scottchiefbaker:
I had a discussion on #perl with the experts and the consensus seems to be that `#!/usr/bin/env perl` is the best. I'm willing to make this change if we can find some documentation somewhere that indicates that `#!/usr/bin/env perl` is "best practice".

I haven't found any *yet*, but I only looked for five minutes. 

@stevemao, @paulirish thoughts?

scottchiefbaker:
This is the closest I could find: 

http://www.cyberciti.biz/tips/finding-bash-perl-python-portably-using-env.html 

but it's hardly "official".

OJFord:
I'd +1 that - I don't think there'll ever be anything better, if it's not in `env` then...

And `#!/usr/bin/env` _is_ Python best-practice (PEP394).

myfreeweb:
You're already using `#!/usr/bin/env bash` ;-)

There's nothing Perl specific about this, it's a common Unix thing.

OJFord:
@myfreeweb Yes, but there could well have been a common or best practice `$PERL_WHICH_FOR_SCRIPT` or something.

scottchiefbaker:
I'm really surprised this isn't a documented best practice somewhere by now.

PR #172, opened 2016-06-15
Correct number of output lines so --patch mode works

scottchiefbaker:
This should address #35


OJFord:
Perhaps you'd noticed, but the "deleted: fname" text is missing between the dashed line header on file deletion in a `--patch`; not in `--patch-mode` when running on a regular `diff`:

screen shot 2016-06-15 at 21 45 45

screen shot 2016-06-15 at 21 46 53


Now I think about it, maybe that was the real cause of the issue addressed by scottchiefbaker/diff-so-fancy#2 - that's one fewer line that expected in the input; 'fixed' by printing an extra at the end.

scottchiefbaker:
Why is the "deleted: fname" not making it in this output? Can you post the raw git diff? I bet if we fix that it will fix this as well.

OJFord:
No idea. I've just looked, and I don't really speak Perl, but `file_change_string` seemingly returns _something_ whatever happens - so the fact that it's empty is a puzzler.

Here's the problem patch as above - the modifcation works; deletion doesn't:
```patch
diff --git a/appveyor.yml b/appveyor.yml
deleted file mode 100644
index a6fb95e..0000000
--- a/appveyor.yml
+++ /dev/null
@@ -1,10 +0,0 @@
-install:
-  - git clone --depth 1 https://github.com/sstephenson/bats.git
-
-build: false
-
-before_test:
-  - git submodule update --init
-
-test_script:
-  - C:\cygwin\bin\bash -lc "cd $APPVEYOR_BUILD_FOLDER; bats/libexec/bats test/*.bats"
diff --git a/diff-so-fancy b/diff-so-fancy
index 0e93e80..8e4dc71 100755
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -26,6 +26,7 @@ print_header_clean () {
   exec "$(get_script_dir)/lib/diff-so-fancy.pl" "$@"
 }
 
+ 
 # run it.
 # shellcheck disable=SC2002
 cat $input \
```

scottchiefbaker:
@OJFord can you give me raw output, I think it may be the ANSI codes messing it up?

    git diff --color | diff-so-fancy | base64

OJFord:
The colouring there is just Github rendering GFM, but sure:
```
ZGlmZiAtLWdpdCBhL2FwcHZleW9yLnltbCBiL2FwcHZleW9yLnltbApkZWxldGVkIGZpbGUgbW9kZSAxMDA2NDQKaW5kZXggYTZmYjk1ZS4uMDAwMDAwMAotLS0gYS9hcHB2ZXlvci55bWwKKysrIC9kZXYvbnVsbApAQCAtMSwxMCArMCwwIEBACi1pbnN0YWxsOgotICAtIGdpdCBjbG9uZSAtLWRlcHRoIDEgaHR0cHM6Ly9naXRodWIuY29tL3NzdGVwaGVuc29uL2JhdHMuZ2l0Ci0KLWJ1aWxkOiBmYWxzZQotCi1iZWZvcmVfdGVzdDoKLSAgLSBnaXQgc3VibW9kdWxlIHVwZGF0ZSAtLWluaXQKLQotdGVzdF9zY3JpcHQ6Ci0gIC0gQzpcY3lnd2luXGJpblxiYXNoIC1sYyAiY2QgJEFQUFZFWU9SX0JVSUxEX0ZPTERFUjsgYmF0cy9saWJleGVjL2JhdHMgdGVzdC8qLmJhdHMiCmRpZmYgLS1naXQgYS9kaWZmLXNvLWZhbmN5IGIvZGlmZi1zby1mYW5jeQppbmRleCAwZTkzZTgwLi44ZTRkYzcxIDEwMDc1NQotLS0gYS9kaWZmLXNvLWZhbmN5CisrKyBiL2RpZmYtc28tZmFuY3kKQEAgLTI2LDYgKzI2LDcgQEAgcHJpbnRfaGVhZGVyX2NsZWFuICgpIHsKICAgZXhlYyAiJChnZXRfc2NyaXB0X2RpcikvbGliL2RpZmYtc28tZmFuY3kucGwiICIkQCIKIH0KIAorIAogIyBydW4gaXQuCiAjIHNoZWxsY2hlY2sgZGlzYWJsZT1TQzIwMDIKIGNhdCAkaW5wdXQgXAo
```

Right - the problem isn't when you run it through d-s-f, it's when you run it thr'ugh `--patch` and d-s-f.

Perhaps it's easiest to replicate like this:
```sh
$ rm appveyor.yml
$ git add -p
```

That gives me the missing line, per first photo above. If instead I do:
```sh
$ rm appveyor.yml
$ git --no-pager diff --no-color | diff-so-fancy --patch-mode
```
it's fine, as per the second.

scottchiefbaker:
I'm unable to recreate the behavior you're seeing:

```
bakers@yellow(~/github/diff-so-fancy)
:cat /tmp/raw.txt | diff-so-fancy --patch-mode
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
deleted: appveyor.yml
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

@ appveyor.yml:3 @
install:

build: false

before_test:
  - git submodule update --init

test_script:
  - C:\cygwin\bin\bash -lc "cd $APPVEYOR_BUILD_FOLDER; bats/libexec/bats test/*.bats"
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
modified: diff-so-fancy
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

@ diff-so-fancy:29 @ print_header_clean () {
  exec "$(get_script_dir)/lib/diff-so-fancy.pl" "$@"
}


# run it.
# shellcheck disable=SC2002
cat $input \
```

OJFord:
Again, per my second photo above, that works fine for me to.

When you actually run it through `git add --patch`, however, it doesn't.

scottchiefbaker:
Hmmmmm can you help me recreate this then? If I `git rm appveyor.yml` and then `git diff --patch` I'm **not** seeing the missing file deleted string.

OJFord:
If you `git rm` you'll need to `diff --cached`.

Assuming that was just a typo here though, I have no idea. I recreated after seeing it in @tomviner's screenshot in #35.

I'm using git v2.9.0 on OS X.

scottchiefbaker:
Maybe it's an OSX only bug? I did exactly what you said above and I can't reproduce it. Maybe someone else with OSX can try and reproduce?

OJFord:
Curious. I changed `diffFilter` to `tee /tmp/dump`, and that still contains the "deleted file" line; so it doesn't seem to be an issue with git.

OJFord:
Oh - d'oh. This whole time it has been printing, just out of order - beneath the `@ fname @` line. I'd seen it there but thought nothing of it.

```sh
~/diff-so-fancy (patch-mode)> git diff --color | ./diff-so-fancy --patch-mode
─────────────────────────────────────────────────────────────────────────────────────────────
deleted: appveyor.yml
─────────────────────────────────────────────────────────────────────────────────────────────

@ appveyor.yml:3 @
install:

build: false

before_test:
  - git submodule update --init

test_script:
  - C:\cygwin\bin\bash -lc "cd $APPVEYOR_BUILD_FOLDER; bats/libexec/bats test/*.bats"

~/diff-so-fancy (patch-mode)> git add -p
─────────────────────────────────────────────────────────────────────────────────────────────
─────────────────────────────────────────────────────────────────────────────────────────────

@ appveyor.yml:3 @
deleted: appveyor.yml
install:

build: false

before_test:
  - git submodule update --init

test_script:
  - C:\cygwin\bin\bash -lc "cd $APPVEYOR_BUILD_FOLDER; bats/libexec/bats test/*.bats"

Stage deletion [y,n,q,a,d,/,?]?
```

Any ideas? I suppose it's not truly out-of-order, since part comes from d-s-f and part from git - so maybe we just need to flush stdout?

OJFord:
But it should be line-buffered.. we're getting closer anyway - and it _is_ out of order, since it comes after the lower half of header.

I don't think I know enough Perl to get much further - but I'm happy to test anything you can think of.

tomviner:
@scottchiefbaker Looking better with latest change.

As you note @OJFord, there's still an ordering bug (deleted: appveyor.yml appears on wrong line, but no longer an error at the end) you can reproduce like this:

```shell
# starting from clean repo
tom:~/dev/diff-so-fancy$ rm appveyor.yml 
tom:~/dev/diff-so-fancy$ git add --patch
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

@ appveyor.yml:3 @
deleted: appveyor.yml
install:
 
build: false
 
before_test:
  - git submodule update --init
 
test_script:
  - C:\cygwin\bin\bash -lc "cd $APPVEYOR_BUILD_FOLDER; bats/libexec/bats test/*.bats"

Stage deletion [y,n,q,a,d,/,?]? n

tom:~/dev/diff-so-fancy$ 
```

scottchiefbaker:
@tomviner AHA! Thanks... now I can recreate it! The raw output that's generating the failure is:

```
diff --git a/lib/diff-so-fancy.pl b/lib/diff-so-fancy.pl
index 3e36905..a7fa90c 100755
--- a/lib/diff-so-fancy.pl
@@ -37,7 +37,10 @@ if ($args =~ /--patch-mode/) {
 my ($file_1,$file_2);
 my $last_file_seen = "";
```

tomviner:
I think I can fix the bad ordering with this extra print: https://github.com/tomviner/diff-so-fancy/commits/patch-mode

now it looks like this

```shell
tom:~/dev/diff-so-fancy$ rm appveyor.yml 
tom:~/dev/diff-so-fancy$ git add --patch
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
deleted: appveyor.yml
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────


@ appveyor.yml:3 @
install:
 
build: false
 
before_test:
  - git submodule update --init
 
test_script:
  - C:\cygwin\bin\bash -lc "cd $APPVEYOR_BUILD_FOLDER; bats/libexec/bats test/*.bats"
Stage deletion [y,n,q,a,d,/,?]? 
```

tomviner:
And I've got another bug, when I make a file executable, I get `Can't open --patch-mode`

```shell
tom:~/dev/diff-so-fancy$ chmod u+x circle.yml 
tom:~/dev/diff-so-fancy$ git add --patch
Can't open --patch-mode: No such file or directory at /usr/local/lib/node_modules/diff-so-fancy/lib/diff-so-fancy.pl line 145.
diff --git a/circle.yml b/circle.yml
old mode 100644
new mode 100755
Stage mode change [y,n,q,a,d,/,?]? n

tom:~/dev/diff-so-fancy$ 
```

scottchiefbaker:
I am able to recreate the bug on Linux (it's not OSX only). I think the problem may be a git bug now though. It appears the output that `git --patch` uses is different than the regular git diff output. Specifically the header is different for deleted files:

A regular diff looks like:

```
diff --git a/appveyor.yml b/appveyor.yml
deleted file mode 100644
index a6fb95e..0000000
--- a/appveyor.yml
+++ /dev/null
@@ -1,10 +0,0 @@
```

but a `git --patch` header is missing a line

```
diff --git a/appveyor.yml b/appveyor.yml
deleted file mode 100644
index a6fb95e..0000000
--- a/appveyor.yml
@@ -1,10 +0,0 @@
```



When d-s-f sees a line that starts with `---` it knows it is the origin file, so it reads the next line to find the destination file. In the case of `diff --patch` the next line is NOT the `+++` destination file line. Which causes all sorts of weirdness. 

Does anyone have a contact on the git team to ask them this if these different behaviors are on purpose or not?

scottchiefbaker:
@tomviner Also... I fixed your executable permission bug and updated the pull request. So we are making SOME progress.

OJFord:
Ah, bingo, good catch.

I think @pef contributed the patch that facilitated this; may have some more info on it.

peff:
> Personally I don't like the +++ /dev/null at all and would be glad to see it gone everywhere - it always strikes me as a "damn we need a second file because of the way we implemented this other thing" hack. But that's beside the point, I assume it's intended to be consistent.

Right, though it's not so much "the way we implemented it" as "that is how `diff` and `patch` have always worked, and git's patches are compatible with those tools". Git added some new meta-information to them, but the fundamentals have to remain, even if they are redundant.

Reading https://github.com/so-fancy/diff-so-fancy/pull/172#issuecomment-226345981, I think the issue is not that the line is missing, but that's reordered.  Here's a simple case (I'm assuming you mean `git add --patch` when you say `git --patch`):

```ShellSession
$ git init -q; echo content >file; git add file; git commit -m foo
$ rm file
$ git diff
diff --git a/file b/file
deleted file mode 100644
index d95f3ad..0000000
--- a/file
+++ /dev/null
@@ -1 +0,0 @@
-content

$ git add --patch
diff --git a/file b/file
index d95f3ad..0000000
--- a/file
+++ /dev/null
deleted file mode 100644
@@ -1 +0,0 @@
-content
Stage deletion [y,n,q,a,d,/,?]? 
```

This seemed eerily familiar, and indeed, digging turned up git/git@e1327ed5fbaca0af44db89d60c33b641e2f21ee1, which was supposed to fix this exact case (jeez, I even used the same example file over 6 years ago). So either that fix didn't work, or it somehow regressed again over the years.

peff:
> This seemed eerily familiar, and indeed, digging turned up git/git@e1327ed, which was supposed to fix this exact case (jeez, I even used the same example file over 6 years ago). So either that fix didn't work, or it somehow regressed again over the years.

Ah, the answer is in the first paragraph of that commit message. That patch is about fixing the earlier dis-ordering we did before feeding the result to `git apply`. But the dis-ordering happens because we want to put "header" stuff at the top, and then "stuff you say yes or no to" separately. And the deletion (along with mode changes) is something in the latter category.

Now as far as diff-so-fancy goes, we should be feeding the original in-order patch to it, and then moving the lines around (in both the real patch and the fancy one) after that. But what I suspect is biting you is that there needs to be a one-to-one correspondence in the non-fancy and fancy lines. That is, if git is moving line 2 of the patch in the example I gave above, then it will move line 2 of the fancy patch, no matter what is in it. So it needs to have a roughly corresponding meaning for the reordered lines to make sense when shown to the user.

OJFord:
> That is, if git is moving line 2 of the patch in the example I gave above, then it will move line 2 of the fancy patch, no matter what is in it. So it needs to have a roughly corresponding meaning for the reordered lines to make sense when shown to the user.

Okay, then if we can assume that `diff --git ...` will not be moved; `+++ ...` and `---` ...` will always come together, I propose something like:
```
diff --git a/file b/file      ->     \n
index d95f3ad..0000000        ->    \n
--- a/file                  { ->    -------------------------------\n deleted: a/file\n
+++ /dev/null               { ->    -------------------------------\n
deleted file mode 100644      ->    \n
```

Then, if I've understood correctly, we'll be shuffling whitespace around (which is unfortunate), but it will never be inside the dashes (which would look worse in my opinion), and `deleted/added/modified` will always be inside. What do you think @scottchiefbaker?

OJFord:
Although perhaps at this point we may as well be just suggesting people use `diff-highlight` as their `diffFilter`.

OJFord:
Also, in re
> it's not so much "the way we implemented it" as "that is how diff and patch have always worked

sorry - I didn't mean that as aggressive as it sounds, and I didn't mean to blame git maintainers either. I meant more like, it stinks of supporting a legacy - or the way in which something (internal or not) is implemented. Though of course, as you point out, it needs to be.

I imagine there wouldn't be many people who, starting from scratch today, would do that though. That's all I meant, really.

Thanks again for the feedback, and for digging out that commit!

peff:
@OJFord No problem. My response was meant to be in "let me expound" mode, not "snippy, well actually..." mode. :wink:

I don't actually use `diff-so-fancy` myself, but I think it's a neat project. Thanks for working on it (I gave it a mention in the Git v2.9 blog post I did recently, which I guess is what spurred some recent bug reports. I hope it didn't catch you guys too off-guard).

scottchiefbaker:
I'm excited to work on this issue, but I'm 100% off the grid for the next two days. I'll be checking emails sporadically, but don't expect much beyond that.

scottchiefbaker:
Where are we at with this issue? Last I checked we were 95% of the way there, but deletes were broken?

OJFord:
My PR scottchiefbaker/diff-so-fancy#3 fixes it as far as I've tested.

So we're pending your review and fixing of my dodgy Perl I suppose :)

henrahmagix:
Since scottchiefbaker/diff-so-fancy#3 has been closed, has there been anything further on this? I split hunks often when using `add --patch` and would love to not get confused about the "off by one" error.

aleksandrs-ledovskis:
Looks like https://github.com/git/git/commit/42f7d45428e260acf535ba0d55ecc91ee81e21da broke d-s-f & `--patch` integration

peff:
For some definition of "break", perhaps. :wink: The point of that patch is to warn people that their patches may be getting corrupted due to the off-by-one problem. So it's detecting and warning on an existing problem.

scottchiefbaker:
@peff would it be possible to add a d-s-f only flag to Git that would make patch mode leave all the lines intact in the original order?

peff:
@scottchiefbaker I think it would have to be a flag to disable hunk-splitting. And then as long as you kept a one-to-one correspondence with hunks, you'd be fine.

peff:
> I think it would have to be a flag to disable hunk-splitting.

Oh, and also the upcoming stage-a-single-line feature. And possibly `e`dit mode, though I think that actually dumps the raw patch into the editor anyway.

scottchiefbaker:
@peff if and when those become reality do you mind letting us know here so we can revisit this issue?

scottchiefbaker:
@peff I didn't realize it was Perl. I just assumed it was C, and I don't know any C.

If you point me in at a good place to start I'll take a look at provided a patch.

OJFord:
> For some definition of "break", perhaps.

It does seem to have broken `diff-highlight` too though @peff (again, for some definition of 'break' 🙂 ) - is this known about?

HaleTom:
Any update?

[This](https://github.com/so-fancy/diff-so-fancy/issues/35#issuecomment-578188120) may be relevant.

Issue #175, opened 2016-06-20
crash on old perl


Chaos99:
diff-so-fancy refuses to work when called e.g. by `git log` and bails out with

```
Possible unintended interpolation of @less_char_vars in string at /usr/bin/lib//diff-so-fancy.pl line 245.
```

This is probably because of my old perl (v5.8.4) on a 32bit machine. It works on a 64bit machine with perl v5.18.2.

Can you give me a hint on a workaround or a patch to still use the script?

scottchiefbaker:
Good catch. I'm not used to writing for super-old Perl's, and that line does indeed have a feature that old Perl's may not have. Perl 5.8 was released in 2002, you must have to _work_ to have a version that old? What OS is it?

As a temporary work around can you change the `//` on line 245 to `||` and see if that fixes the problem? If it does I'll fix the core code so we're Perl 5.8.x compliant.

Chaos99:
Well, I wouldn't have dared to call this a bug or demand a patch. I very well know that my perl is ancient and this was just a kind request for a hint or workaround. So thanks for the quite positive response so far.

The Heisenbergian work/shame relation applies here. I can either tell you were I work or how old my system is, but never both. As this is my private account: I'm working within a Debian 3.1 sarge, gcc 3.3.5 and glibc 2.1.3. My Python is 2.3.5. As you can imagine, my ~/local folder is quite huge and I'm running half the system from self-compiled parallel versions that at least halfway follow current releases. 
As I don't use perl much, I really DON'T want to wade through the whole dependency tree to also parallel-install that. 
So thanks very much for the advice, I will report back tomorrow.

scottchiefbaker:
If at-all-possible we will work to make d-s-f as compatible as possible. As I'm sure you understand there are limits to what we can do, but in this case it seems like there is a semi-simple work around. Assuming this works for you I'm more than happy to commit that to `master` and solve the problem for you.

Chaos99:
Works with the change on both of my systems.

 Thanks a lot.

scottchiefbaker:
Fix has landed in the `master` branch

PR #177, opened 2016-06-23
Fix for old Perl 5.8

delphinus35:
The diff in #175 is insufficient for Perl 5.8.  This patches additional `//` => `||` and escaping `@` in regexp.

### DIFF
--- a/lib/diff-so-fancy.pl
+++ b/lib/diff-so-fancy.pl
@@ -160,7 +160,7 @@ while (my $line = <>) {
 # Courtesy of github.com/git/git/blob/ab5d01a/git-add--interactive.perl#L798-L805
 sub parse_hunk_header {
 	my ($line) = @_;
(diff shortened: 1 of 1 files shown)


scottchiefbaker:
I had to consult the gurus on #perl on the escaped @ in the regexp. It's not required, but it will add clarity. This looks good to me. Merging.

Thanks for the contribution.

delphinus35:
Not `@` but `@+` is the problem.  `@+` means [`@LAST_MATCH_END`][lme] and this odd name variable cannot be use in regexp [before 5.10][510].

So actually, it is only needed that one `\` escaping such as `@@+` → `@\@+`.  I have added in excess ;)

[lme]: http://perldoc.perl.org/perlvar.html#%40LAST_MATCH_END
[510]: http://perldoc.perl.org/perl5100delta.html#%40--and-%40%2b-in-patterns

PR #179, opened 2016-06-26
clarify how to add an alias in readme

### LINKED ISSUE #178: gitconfig error: adding dsf alias to gitconfig causes error
Adding a `dsf` alias to `~/.gitconfig` as per Readme causes error when running Git commands.

Env: OS X, Bash
```
[user]
        email = someone@example.com
[core]
        excludesfile = ~/.gitignore

### diff-so-fancy
git config --global alias.dsf '!f() { [ -z "$GIT_PREFIX" ] || cd "$GIT_PREFIX" && git diff --color "$@" | diff-so-fancy | less --tabs=4 -RFX; }; f'
```
>$ git status
>fatal: bad config file line 7 in /Users/zzz/.gitconfig

Appreciate any pointers to make this work. Thanks.

stevemao:
fixes #178

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -24,7 +24,7 @@ git diff --color | diff-so-fancy
 git config --global core.pager "diff-so-fancy | less --tabs=4 -RFX"
 ```
 
-However, if you'd prefer to do the fanciness on-demand with `git dsf`, add an alias to your `~/.gitconfig`
+However, if you'd prefer to do the fanciness on-demand with `git dsf`, add an alias to your `~/.gitconfig` by running:
 ```shell
 git config --global alias.dsf '!f() { [ -z "$GIT_PREFIX" ] || cd "$GIT_PREFIX" && git diff --color "$@" | diff-so-fancy | less --tabs=4 -RFX; }; f'
 ```
--- a/readme.md
+++ b/readme.md
@@ -26,7 +26,8 @@ git config --global core.pager "diff-so-fancy | less --tabs=4 -RFX"
 
 However, if you'd prefer to do the fanciness on-demand with `git dsf`, add an alias to your `~/.gitconfig` by running:
 ```shell
-git config --global alias.dsf '!f() { [ -z "$GIT_PREFIX" ] || cd "$GIT_PREFIX" && git diff --color "$@" | diff-so-fancy | less --tabs=4 -RFX; }; f'
+git config --global alias.dsf '!f() { [ -z "$GIT_PREFIX" ] || cd "$GIT_PREFIX" '\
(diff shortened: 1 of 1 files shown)


scottchiefbaker:
This is a good clarification. I wonder if we shouldn't break it up with some `\` on the end so it's more readable. It's a pretty gnarly long scroll to the right.

stevemao:
Sounds good. @scottchiefbaker maybe add some comments on what it does too.

scottchiefbaker:
What about something like this:

```
git config --global alias.dsf '!f() { [ -z "$GIT_PREFIX" ] || '\
'cd "$GIT_PREFIX" && git diff --color "$@" | diff-so-fancy  | '\
'less --tabs=4 -RFX; }; f'
```

scottchiefbaker:
Or maybe just two lines, breaking at char 82 instead?

```
git config --global alias.dsf '!f() { [ -z "$GIT_PREFIX" ] || cd "$GIT_PREFIX" '\
' && git diff --color "$@" | diff-so-fancy  | less --tabs=4 -RFX; }; f'
```

I just don't want a scrollbox on the Github page. Make it harder to read and copy/paste.

paulirish:
how about

```sh
DSF_CMD='!f() { [ -z "$GIT_PREFIX" ] || cd "$GIT_PREFIX" && git diff --color "$@" | diff-so-fancy | less --tabs=4 -RFX; }; f'
git config --global alias.dsf $DSF_CMD
```

scottchiefbaker:
@paulirish that's a good start, but it's still 126 chars wide (and has a scroll bar on my screen). Alternately we could do:

```
DSF_CMD='!f() { [ -z "$GIT_PREFIX" ] || cd "$GIT_PREFIX" &&'
DSF_CMD+=' git diff --color "$@" | diff-so-fancy | less --tabs=4 -RFX; }; f'

git config --global alias.dsf $DSF_CMD
```

Not sure if this is better syntax than a bash `\` though.

stevemao:
I don't like the `DSF_CMD+=` part. I like your previous suggestions better @scottchiefbaker :)

scottchiefbaker:
Which one should we go with? Two or three line `\` version?

stevemao:
I went with your first option @scottchiefbaker

scottchiefbaker:
Thanks for the vote of confidence. I **just** realized there is an extra space in my command (it works fine though). We should probably update this to:

```
git config --global alias.dsf '!f() { [ -z "$GIT_PREFIX" ] || cd "$GIT_PREFIX" '\
'&& git diff --color "$@" | diff-so-fancy  | less --tabs=4 -RFX; }; f'
```

Note the removed space on the second line before the `&&`. I'm totally fine with merging this if we can remove the extra space I screwed up on.

PR #180, opened 2016-07-06
Update recommended pre-seeded search pattern

woodrowbarlow:
The old recommended pre-seeded search pattern for less would cause "git
show" to jump past the commit message, going straight to the file. This
change would prevent that (mis)behavior, without hindering the
pattern's usefulness.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -111,7 +111,7 @@ git --no-pager diff
 
 #### Moving around in the diff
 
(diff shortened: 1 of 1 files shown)


woodrowbarlow:
I've just noticed this has the unintended (although not necessarily *bad*) side-effect of causing "git log" to jump to the first instance of "Date: " (which is at the top of the first entry.

I'm keeping it in my config, because it means I can jump through commit messages in my git log more efficiently, but it is worth mentioning here in case this is undesirable as a default recommendation.

PR #185, opened 2016-07-15
Add setting to allow disabling of unicode ruler

thenoseman:
This commit adds a single setting to allow users
to disable the use of the unicode em-dash that is being used
as a horizonal ruler character for marking the changed filename.

It defaults to the current behaviour of using em-dash.

Setting `diff-so-fancy.useUnicodeRuler` to `false` will use a simple
dash (minus sign) instead.

This fixes a problem I have when using diff-so-fancy with my iTerm2 settings.
In my case the ruler would span multiple lines:
diff-so-fancy-broken-iterm2

Cheers!

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -84,12 +84,18 @@ Simplify git header chunks to a more human readable format.
 
 Should the pesky `+` or `-` at line-start be removed.
 
+### useUnicodeRuler
+
+Should the horizonal line above and below the name of the changed file be displayed using unicode.  
+Try setting this to `false` if those lines are displayed to long to fit your screen.
+
 By default all the configs are true. You can turn any off by running:
 
 ```
 git config --bool --global diff-so-fancy.markEmptyLines false
 git config --bool --global diff-so-fancy.changeHunkIndicators false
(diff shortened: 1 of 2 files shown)


readme.md
```diff
@@ -84,12 +84,18 @@ Simplify git header chunks to a more human readable format.
 
 Should the pesky `+` or `-` at line-start be removed.
 
+### useUnicodeRuler
+
+Should the horizonal line above and below the name of the changed file be displayed using unicode.
```
scottchiefbaker: I'd like to change the wording on this to something like:

```
By default the separator for the file header uses unicode line drawing characters. If this is 
causing output errors on your terminal set this to false to use ASCII characters instead.
```

scottchiefbaker:
I like this a lot actually. I'd like to change two lines in the documentation to be slightly more readable, and then I'll merge this.

+1

thenoseman:
Go ahead. I'm totally fine with it.

scottchiefbaker:
Please update the pull request with the documentation rewording above and I'll merge this pull request.

PR #188, opened 2016-07-29
Improve "pro-tips" documentation

### LINKED ISSUE #173: Pro-tip less with core.pager configuration results in "Pattern not found (press RETURN)" for other commands
When using the `core.pager` configuration as specified in the README with the "Pro-tip" of pre-seeding a search pattern with less, other commands such as `git log` show:

`Pattern not found  (press RETURN)`

Once enter is hit, the normal log output is shown. The documentation should probably suggest a better way of doing this, or not suggest it at all.

OJFord:
- Explicitly limits suggestion to over-eager load a search for navigation purposes to diffs (closes #173).
- Adds a new tip for `patch` alias to render DSF-less `.patch`es.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -117,6 +117,16 @@ Sometimes you will want to bypass diff-so-fancy. Use `--no-pager` for that:
 git --no-pager diff
 ```
 
+#### Raw patches
+
+As a shortcut for a 'normal' diff to save as a patch for emailing or later
+application, it may be helpful to configure an alias:
+```ini
+[alias]
+    patch = --no-pager diff --no-color
+```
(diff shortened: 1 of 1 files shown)


stevemao:
I don't mind adding this if everyone's happy :) One thing I wanna point out is I don't want users to treat this as a shell or git reference.

scottchiefbaker:
I'm totally fine with this. +1 for pro-tips.

OJFord:
@stevemao Agreed on not being a shell or git reference, but I think more examples help users understand how dsf fits in with other tooling; reduces the "I have no idea what this is doing, I just dropped the line they told me to in my `.gitconfig`" effect.

PR #189, opened 2016-08-02
Change `less` options to be mere suggestion

### LINKED ISSUE #174: Why --tabs=4 for less?
As per subject. I'd like to use this tool without entering a religion war on the size of tabs 😄.
- Either `--tabs=4` is used only because the authors prefer so for other reasons, and is not actually needed. and then the README could be improved;
- Or diff-so-fancy actually relies on that, and then it might not be compatible with some codebases I'm using that rely on tabs being 8 space wide.

My current config:
```
$ git config --global core.pager
less -+$LESS -FRX -i
```

Config I plan to try out:
```
git config --global core.pager='diff-so-fancy |less -+$LESS -FRX -i'
```

OJFord:
Closes #174.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -19,7 +19,7 @@ You can do one-off fanciness:
 git diff --color | diff-so-fancy
 ```
 
-**But**, you'll probably want to fancify all your diffs. Run this so `git diff` will use it:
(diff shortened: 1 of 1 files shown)


scottchiefbaker:
Works for me!

PR #190, opened 2016-08-06
Disable fancification when redirecting output to a file

### LINKED ISSUE #181: Disable dsf when piping to file?
I have used the suggested git pager replacement from the readme (`git config --global core.pager "diff-so-fancy | less --tabs=4 -RFX"`) so that commands like "git diff" and "git show" use diff-so-fancy by default. Is is possible to write this config option such that diff-so-fancy is only used when the output stream is stdout?

Basically, I want to be able to do `git diff HEAD^..HEAD > changes.patch` and end up with a usable patch file, whereas currently I need to do `git --no-pager diff --no-color HEAD^..HEAD > changes.patch`.

OJFord:
This commit aims to better support standard patch files, by
disabling colours and header removal when the stdout file
descriptor is opened on a terminal.

The assumption is that if this is _not_ the case, it is most
likely that the diff output is being redirected to a .patch file,
for emailing or some other medium in which fancified output may
be confusing, and not compatible with standard tools.

We do this by internalising `less`, providing the bare minimum
`--RAW-CONTROL-CHARS` option required for diff-so-fancy to operate
as expected, and appending any user-supplied options.

~~This is actually _NOT_ a breaking change.~~ \* `less` performs a
similar test, and exits leaving its input on stdout. This means
that a user who upgrades but keeps `diff-so-fancy | less $OPTS`
will ~~still see the new behaviour; keeping~~ keep their existing `less`
configuration. (The `-R` requirement is not new; this hypothetical
user must have had at least `OPTS=R`.)

This commit closes #181.
- See [my comment below](https://github.com/so-fancy/diff-so-fancy/pull/190#issuecomment-238027844). It's not breaking, for now, but I think it would be a good idea to - for reasons outlined in 5582a58.

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
 [[ $# -ge 1 ]] && [[ -f "$1" ]] && input="$1" || input="-"
+[[ $# -ge 1 ]] && less_opts="${*:1}" || less_opts=
 
 get_script_dir () {
   src="${BASH_SOURCE[0]}"
@@ -26,8 +27,11 @@ print_header_clean () {
   "$(get_script_dir)/lib/diff-so-fancy.pl"
 }
 
-# run it.
-# shellcheck disable=SC2002
-cat $input \
-  | $diff_highlight \
-  | print_header_clean
+if [[ -t 1 ]]; then
+    # run it.
+    # shellcheck disable=SC2002
+    cat $input | $diff_highlight | print_header_clean | less -R $less_opts
+else
+    # stdout is not going to a terminal; probably a patch file
+    cat $input
+fi
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -27,11 +27,13 @@ print_header_clean () {
   "$(get_script_dir)/lib/diff-so-fancy.pl"
 }
 
-if [[ -t 1 ]]; then
+FD_is_pipe=$(perl -e 'exit(-p STDOUT ? 0 : 1);')
+FD_is_terminal=$(test -t 1)
+if $FD_is_terminal || $FD_is_pipe; then
     # run it.
     # shellcheck disable=SC2002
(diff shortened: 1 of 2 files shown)


OJFord:
Hah, d'oh. Tests fail because when they run, of course stdout isn't sent to a terminal; so d-s-f disables itself.

I'll have to have a think about how to test it; in the mean time I'll pretend that this was a deliberate demonstration of the feature working 😉

stevemao:
> This is actually NOT a breaking change. less performs a
> similar test, and exits leaving its input on stdout.

This means people using other pagers such as vim would still work?

OJFord:
> This means people using other pagers such as vim would still work?

I typo'd pretty badly there. They would _not_ see the new behaviour, but nothing would break.

Yes, other pagers such as vim would continue to work - but again, they would not get the benefit of the new behaviour.

OJFord:
Actually, it's worse than that. I was completely wrong when I wrote that, it does break; of course it does - for everybody.

I'll have another look at this later. I'm now thinking it might be best to have an optional `--pager=whatever` argument and `--pager-opts="--whatever -y -o -u --would-like"`.

OJFord:
Okay, I fixed that so that it continues to work with vim, or anything else @stevemao.

My suggestion is:
1. The standard use becomes simply `diff-so-fancy`, which internally uses `less -R`
2. I'll add a method of setting a custom pager, so that this can be changed to `vim -`, or `less -rFX --tabs=4`, or anything
3. Release a minor version with a deprecation warning for explicitly piping to a pager
4. Enjoy the passage of time
5. Release a major version which reverts 5582a58 - explicitly piping unsupported

What do you think?

scottchiefbaker:
What is the 2017 status of this?

OJFord:
As we discussed in #181, it seems the main goal of this PR have been replicated elsewhere and made it to master, so I guess it can be closed.

Should be okay as of v0.11.3.

PR #191, opened 2016-08-08
change lib/ to libexec/

stevemao:
fixes https://github.com/Homebrew/homebrew-core/pull/3663#issuecomment-238010069


scottchiefbaker:
Interesting.... works for me.

PR #206, opened 2016-11-19
Addition of Arch linux package availability

chrissound:


scottchiefbaker:
I have no problem with this commit... but I'm wondering if we shouldn't move it to a section at the bottom for Linux installs. Each distro is going to have it's own slightly different way to do the same thing.

stevemao:
Thanks @scottchiefbaker :) I'm happy to merge this for now and reorganise sections in the future (when there are more linux installs etc). But anyone wanna reorganise now feel free :)

PR #212, opened 2017-01-13
Fix README patch alias

Kmaschta:
When I set `patch = --no-pager diff --no-color` in my gitconfig and type `git patch`, I have this error:

```
fatal: alias 'patch' changes environment variables
You can use '!git' in the alias to do this.
```

Then, here is a fix.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -128,7 +128,7 @@ As a shortcut for a 'normal' diff to save as a patch for emailing or later
 application, it may be helpful to configure an alias:
 ```ini
 [alias]
-    patch = --no-pager diff --no-color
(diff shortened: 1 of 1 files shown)


OJFord:
LGTM.

OJFord:
ping @scottchiefbaker

scottchiefbaker:
I don't use d-s-f for patches. If you're ok with it (this is only a documentation thing) then I'm ok with it.

OJFord:
Yep, as it was it would have given an error if someone just copy-pasted into their `.git/config` without checking (or knowing it was wrong).

PR #214, opened 2017-01-20
Use /usr/bin/env for perl

raviqqe:


scottchiefbaker:
I'm fine with this merge, as this has come up before.

I *think* it's best practice to use `/usr/bin/env` but I'd like to see it in writing before we merge this. `/usr/bin/perl/` is pretty well established, so before we make this change I want to see some research.

OJFord:
Yep, we discussed it in #171.

[You said](https://github.com/so-fancy/diff-so-fancy/pull/171#issuecomment-226036186):

> I had a discussion on #perl with the experts and the consensus seems to be that #!/usr/bin/env perl is the best

🙂

scottchiefbaker:
Heh... so I did :)

raviqqe:
diff-so-fancy is working well on my FreeBSD box now. Thanks.

PR #193, opened 2016-08-15
Mercurial support

frantic:
This will help with Mercurial compatibility – it doesn't output `index` line, and `diff-so-fancy` depends on it to draw the first horizontal line.

All `bats test`s pass.

Fixes #186

PR #218, opened 2017-02-18
Add fancy install script, closes #217

### LINKED ISSUE #217: Install using a curlable bash script
Hello folks!

Great job with this nice little tool. I was wondering if there is a way to curl in and execute an sh script, like [the nvm install script](https://github.com/creationix/nvm#install-script).

Thanks a lot.

arichiardi:
Here it is, can of course be improved on fanciness :smoking:

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,164 @@
+#!/usr/bin/env bash
+#
+# diff-so-fancy install script
+#
+# diff-so-fancy builds on the good-lookin' output of git contrib's
+# diff-highlight to upgrade your diffs' appearances.
+#
+# Output will not be in standard patch format, but will be readable.
+# No pesky + or - at line-start, making for easier copy-paste.
+#
+# largely inspired by nvm's install script
+
+GREEN=`tput setaf 2`
+RED=`tput setaf 1`
+NC=`tput sgr0` # No Color
+UNDERLINE=`tput smul`
+BOLD=`tput bold`
+INSTALL_DIR=
+
+{ # this ensures the entire script is downloaded #
+
+dsf_has() {
+    type "$1" > /dev/null 2>&1
+}
+
+dsf_latest_version() {
+    echo "v0.11.4"
+}
+
+dsf_download() {
+    if dsf_has "curl"; then
+        curl -q "$@"
+    elif dsf_has "wget"; then
+        # Emulate curl with wget
+        ARGS=$(echo "$*" | command sed -e 's/--progress-bar /--progress=bar /' \
+                                   -e 's/-L //' \
+                                   -e 's/-I /--server-response /' \
+                                   -e 's/-s /-q /' \
+                                   -e 's/-o /-O /' \
+                                   -e 's/-C - /-c /')
+        # shellcheck disable=SC2086
+        eval wget $ARGS
+    fi
+}
+
+dsf_download_file() {
+    local source_url="https://raw.githubusercontent.com/so-fancy/diff-so-fancy/$(dsf_latest_version)"
+    local param_array=("$@")
+
+    # Array content
+    # 0 source file (will be concatenated to the url)
+    # 1 destination file (used as passed in)
+    # 2 string that if equal to "exec" will set chmod +x
+    local param_src="${param_array[0]}"
+    local param_dst="${param_array[1]}"
+    local param_mod="${param_array[2]}"
+
+    echo -en "${BOLD}Download $param_src: ${NC}"
+
+    dsf_download -s "$source_url/$param_src" -o "$param_dst" || {
+        echo -e "${BOLD}${RED}KO${NC}"
+        echo -e >&2 "${BOLD}${RED}Failed to download '$source_url/$param_src' to '$param_dst'${NC}\nPlease try again."
+        exit 2
+    }
+    if [ "exec" = "$param_mod" ]; then
+        chmod a+x "$param_dst" || {
+            echo -e >&2 "${RED}Failed to mark '$param_dst' as executable${NC}"
+            exit 3
+        }
+    fi
+    echo -e "${BOLD}${GREEN}OK${NC}"
+}
+
+dsf_install_dir() {
+    local dirs=("$HOME/bin" "/usr/local/bin");
+    local writable=
+    local in_path=
+
+    for d in "${dirs[@]}"; do
+        in_path=
+        writable=
+
+        echo -en "${BOLD}Check if '$d' is writable: "
+
+        if [[ -w "$d" ]]; then
+            echo -e "${GREEN}OK${NC}"
+            writable="yes"
+        else
+            echo -e "${RED}KO${NC}"
+        fi
+
+        echo -en "${BOLD}Check if '$d' is in PATH: "
+
+        if [[ "$PATH" =~ $d ]]; then
+            echo -e "${GREEN}OK${NC}"
+            in_path="yes"
+        else
+            echo -e "${RED}KO${NC}"
+        fi
+
+        if [[ "$writable" == "yes" && "$in_path" == "yes" ]]; then
+            INSTALL_DIR=$d
+            break
+        fi
+    done
+
+    if [ -z "$INSTALL_DIR" ]; then
+        echo -e >&2 "${BOLD}${RED}The default install directories could not be used.${NC}\nTry to fix the problem or call the script with sudo."
+        exit 4
+    fi
+}
+
+dsf_install_as_script() {
+    local files1=("diff-so-fancy" "$INSTALL_DIR/diff-so-fancy" "exec")
+    local files2=("third_party/diff-highlight/diff-highlight" "$INSTALL_DIR/diff-highlight" "exec")
+
+    # Checking if it has already been installed
+    if [ -f "$INSTALL_DIR/${files1[0]}" ]; then
+        echo -e >&2 "${BOLD}${RED}The diff-so-fancy script is already installed in '$INSTALL_DIR'${NC}.\nTry to delete '$INSTALL_DIR/${files1[0]}' and try again.${NC}"
+        exit 1
+    else
+        dsf_download_file "${files1[@]}"
+        dsf_download_file "${files2[@]}"
+    fi
+}
+
+dsf_do_install() {
+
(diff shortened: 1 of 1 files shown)


arichiardi:
Still not working, but close :wink:

arichiardi:
Works now! Let me know or push a commit if there are issues ok?

scottchiefbaker:
I like the idea of this, but I think it needs a couple of tweaks

1. Some sort of error if the user doesn't have git installed
2. Error out if ~/bin/ isn't in the path
    1. Offer an option to install to another directory in path?

Overall I'm impressed though. Good work!

arichiardi:
Oh oh well, I guess I gave for granted that they would have git 😀

What the error text could be? 

About the  bin folder, this is actually a user preferences probably...

OJFord:
> Some sort of error if the user doesn't have git installed

Is that really necessary?

The install process doesn't depend on git, #193 is adding Mercurial support, and we're even discussing in #220 how to use dsf with non-VCS diffs.

Even if I am using dsf with git, if I have my own install script which does `... && install_dsf && install_git && ...`, it's annoying if I have to change that after discovering it errors because dsf needs git at _runtime_ (though it actually doesn't, per above) if it doesn't at install time.

arichiardi:
Ah good to know,.

Also, where do folks usually install bin? I personally use `$HOME/bin`, but `usr/local/bin` is probably the classic Linux default.

OJFord:
> Also, where do folks usually install bin? I personally use $HOME/bin, but usr/local/bin is probably the classic Linux default.

Personally I think `/usr/local/bin` is better, pretty likely to already be on `$PATH` whereas `$HOME/bin` will be only for you, and others that deliberately use it. I do too, but only for small '`n`-off' (for `n` slightly larger than `1`) stuff that I've written for myself - so I'd be _slightly_ annoyed if an install script put something third-party there. But that might just be me.

I certainly don't think `~/bin` not already on `$PATH` should be an error. That will hit the vast majority of users, and they'd be surprised.

It would also mean this script couldn't be used in the Homebrew formula - which should IMO be a goal, (not that what I think matters - up to @stevemao :slightly_smiling_face:) at the moment it adds a node/npm dependency, which is not necessary but presumably done for ease of maintanence, which this script can hopefully solve. It'd be nice not to have `brew` install all of node just for a relatively tiny bash/perl script.

arichiardi:
@OJFord so can you confirm that `brew` installs in `/usr/local/bin`? Or shall I pass a parameter there? Not very familiar with `brew`...

scottchiefbaker:
I forgot about mercurial/vanilla diff support. You can ignore my comment a missing git error message.

My vote for installation location is `~/bin/` **if** it's available/created. By default a lot of installs won't have this, A fall back to `/usr/local/bin/`, would be my second choice, but that will require root to write to.

scottchiefbaker:
Think the installer should attempt to put the files in `~/bin/` **IF** it's already in the path. If it's not in the path attempt to put the files in `/usr/local/bin` but error out if the user is not root. That should cover both our install scenarios.

Perhaps this is a separate issue, but I ***HATE*** the default colors. After install people will expect the output to be the same as the screenshot in the README. Perhaps we set the colors for the user? Or offer an option to use the d-s-f "suggested colors"?

arichiardi:
If we have a way to identify if we are being executed in `brew` this can be conditional as well.
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

OJFord:
> can you confirm that brew installs in `/usr/local/bin`

I can.

> Perhaps we set the colors for the user?

Please don't do this by default. Installing dsf shouldn't alter my non-dsf config options unless I explicitly opt in; colours are `diff` (git core) and `diff-highlight` options.

scottchiefbaker:
> Please don't do this by default. Installing dsf shouldn't alter my non-dsf config options unless I explicitly opt in; colours are diff (git core) and diff-highlight options.

We should definitely make it an option... "Use default d-s-f colors (will update your gitconfig)?" or something like that.

scottchiefbaker:
@arichiardi where are we at on this? I'm ready to merge this if we can include an option to set the default colors.

```
[color]
	ui = always
[color "diff"]
	meta = yellow bold
	commit = green bold
	frag = magenta bold
	old = red bold
	new = green bold
	whitespace = red reverse
[color "diff-highlight"]
	oldNormal = red bold
	oldHighlight = "red bold 52"
	newNormal = "green bold"
        newHighlight = "green bold 22"
```

arichiardi:
@scottchiefbaker oh cool!
I have had a pretty busy open source month :smile: Feel free to add this in and a subsequent commit can add default colors.

scottchiefbaker:
@arichiardi we're really close to a 1.0.0 release of d-s-f and I really want this to be part of that. Will you have any time coming up that we can refresh this with a couple of tweaks and then I can land it?

arichiardi:
Yes, I can work on it, let me know when you think changes are stable for me
to start tweaking :)

scottchiefbaker:
`master` is stable right now, so you can make changes as you see fit. As the *last* part of the install script I'd like to offer to set the default colors for the user. There is a new option:

`diif-so-fancy --colors`

That will output the commands needed to set the default colors. Perhaps we can include that in some fashion.

scottchiefbaker:
As of now the only files that are needed are `/diff-so-fancy` and `/third_party/diff-highlight/diff-highlight`.

The `libexec` directory is no longer needed.

scottchiefbaker:
Actually for simplicity you could put both `diff-so-fancy` and `diff-highlight` in `~/bin` (or anywhere in the path) and they will work just fine. 

No need for a `/third_party` directory at all.

arichiardi:
True, will do that

arichiardi:
Done! :fireworks: :tada: :balloon:

scottchiefbaker:
This was with a brand new vanilla user on Fedora 25:

```
test@basement(~)
:sh install.sh
Check if '/home/test/bin' is writable: KO
Check if '/home/test/bin' is in PATH: OK
Check if '/usr/local/bin' is writable: KO
Check if '/usr/local/bin' is in PATH: OK
The default install directories could not be used.
```

In the first case, it looks like `~/bin` is in the path by default, but not created? In that case I think it's a good idea to create the directory and put the files there. This was the behavior of the previous, and I think it's fine **if** `~/bin` is in the path.

Lastly I'd like the install script to offer to set the recommended colors. `diff-so-fancy --colors` will output the commands for the recommended colors.

scottchiefbaker:
Moving forward, I think this will be the recommended way to install d-s-f. Rather than cloning the whole repo, or using sytem repos. This will always ensure that the user gets the latest version via the `master` branch.

arichiardi:
Yes I removed the creation part, assuming that if it is in the path, then it is writable/exists. But of course this assumption was wrong as it was shown above. I can bring it back to life.

scottchiefbaker:
Yes please bring that logic back to life, if the directory is in the path and writable.

arichiardi:
No well, if the directory is writable it means that it exists already...so only if it is in the path. Will do that.

scottchiefbaker:
Question...

Do we want to install to `~/bin/` by default... or require the use to specify?

```
bash install.sh /path/to/install/to/
```

Or both... if you call `install.sh` without a path it tries `~/bin` and errors if it can't?

arichiardi:
Yeah that is a UX question, I am fine with the current defaults, but I am going to wait for more answers to this one

scottchiefbaker:
@paulirish @stevemao @OJFord 

Feedback on this?

install.sh
```diff
@@ -0,0 +1,164 @@
...
+    else
+        dsf_download_file "${files1[@]}"
+        dsf_download_file "${files2[@]}"
+    fi
+}
+
+dsf_do_install() {
+
+    if ! dsf_has dsf_download; then
+        echo >&2 "You need curl or wget to install nvm";
```
paulirish: s/nvm/diff-so-fancy

paulirish:
I really love the `~/bin/` convention and suggest keeping that. If someone doesn't want an easy install script, then they should be fine with our _Manual Install_ directions.

It seems reasonable for this script to doublecheck that ~/bin is in your PATH though. Adding it to the PATH should definitely stay the user's job. Hate when scripts toss stuff into my .bash_profile. :)

arichiardi:
@paulirish What about the creation of the folders if already in `PATH`? I usually don't like when scripts are messing with my stuff as well, but I assume that if you have something in `PATH` you actually want/wanted to have the folder in place and therefore I create it (maybe again if it had been removed).

On the other end if a user delete the folder but forgot to update `PATH`, creating it will maybe make her upset :smile:

paulirish:
oh i see you already do the PATH check. :) all good.

i think the way you currently handle it is best (is it writeable? is it in path? great, lets go)

scottchiefbaker:
@paulirish I'd like to completely revamp the README once we release 1.0.

It's a jumbled mess, with a ton of options.

paulirish:
@scottchiefbaker oh word. yeah that sounds excellent! though i'd advocate for the revamp _before_ shipping 1.0, assuming the release brings us some new attention. :)

arichiardi:
Ups pushed the wrong version wait a sec

arichiardi:
Ok pushed the right version, still uncertain about reintroducing `mkdir` or not. Seeking advice.

arichiardi:
What is the status on this? Is there anything I need to add/remove so that we can merge it? Good job with the release!

scottchiefbaker:
I'm not sure where to go with this, I'm open to suggestions. With the release of 1.1.0 the installation is much simpler (one script). I think there is still a place for an install script that downloads the latest version and puts it in your $PATH appropriately. What do you think?

OJFord:
> an install script that downloads the latest version and puts it in your $PATH appropriately. What do you think?

Sounds like a package manager?

arichiardi:
Oh well if there is one script already for that, and easy to call, then there is no need for an additional script.

scottchiefbaker:
Now that things are really simple, I don't think we need a full blown install script. Maybe we just need to expand the install docs to cover some other use cases? I'm going to close this for now. If there are strong objections let me know and I'll re-open it.

PR #222, opened 2017-02-28
Respect if a user had diff.noprefix set in their config (fixes #205)

### LINKED ISSUE #205: Files in a/ directory miss their directory prefix
I've got a change in a file called `a/index.php`.

With `git --no-pager diff` I get:

```
diff --git a/index.php a/index.php
```

dsf shows:

```
modified: index.php
```

scottchiefbaker:


scottchiefbaker:
Paging: @OJFord @stevemao @lode 

Thoughts or concerns before I land this?

libexec/diff-so-fancy.pl
```diff
@@ -18,6 +18,7 @@
 my $strip_leading_indicators   = git_config_boolean("diff-so-fancy.stripLeadingSymbols","true");
 my $mark_empty_lines           = git_config_boolean("diff-so-fancy.markEmptyLines","true");
 my $use_unicode_dash_for_ruler = git_config_boolean("diff-so-fancy.useUnicodeRuler","true");
+my $diff_no_prefix             = git_config_boolean("diff.noprefix","false");
```
OJFord: Seems like this variable actually represents "diff not no prefix" if you see what I mean - perhaps `$diff_show_prefix` would express it better?

OJFord:
Doesn't seem to work for me, am I missing something?

```sh
~/diff-so-fancy/(noprefix|●1)>git config diff.noprefix false
~/diff-so-fancy/(noprefix|●1)>git --no-pager diff --cached
diff --git a/appveyor.yml b/tst.yml
similarity index 94%
rename from appveyor.yml
rename to tst.yml
index a6fb95e..f481b08 100644
--- a/appveyor.yml
+++ b/tst.yml
@@ -1,7 +1,6 @@
 install:
-build: false
 before_test:
   - git submodule update --init
~/diff-so-fancy/(noprefix|●1)>git --no-pager diff --cached | ./diff-so-fancy.pl
similarity index 94%
rename from appveyor.yml
rename to tst.yml
────────────────────────────────────────────────────────────────────────────────
renamed: appveyor.yml to tst.yml
────────────────────────────────────────────────────────────────────────────────
@ tst.yml:3 @
install:
build: false
before_test:
  - git submodule update --init
~/diff-so-fancy/(noprefix|●1)>
```

libexec/diff-so-fancy.pl
scottchiefbaker: I was trying to recreate the exact name from the gitconfig... if we did `$diff_show_prefix` it reverses the logic. I see what you mean, but I'd like to keep it the same name as the gitconfig.

Usually I'm not a fan of negative variable names, but this is the one exception because we're using an existing name.

scottchiefbaker:
You need to set `git config diff.noprefix true` ... false is the default and does **not** strip the `a/` and `b/`.

libexec/diff-so-fancy.pl
OJFord: Okay, but in that case the value returned from `git_config_boolean` should be negated I believe? Do you see what I'm getting at - it seems like `noprefix = false` => `$diff_no_prefix = true` currently.

OJFord:
> false is the default and does not strip the `a/` and `b/`.

Yep, but see the output above, `a/` and `b/` have been stripped when I pipe it through dsf on this branch?

scottchiefbaker:
No I'm not following:

scottchiefbaker:
No I'm not following. From you above example:

```
index a6fb95e..f481b08 100644
--- a/appveyor.yml
+++ b/tst.yml
```

That looks to me like they (`a/` and `b/`) were **not** stripped, which is the correct behavior when you specify false. Unless I'm mis-reading your example.

libexec/diff-so-fancy.pl
scottchiefbaker: Ok I see what you're saying. I can change this to `$git_strip_prefix` to be 100% clear.

scottchiefbaker:
Just so I'm being 100% clear... when you set: 

```
git config diff.noprefix true
```

This cause git to **NOT** output the fake `a/` and `b/` directory prefixes on file names. Setting it to false (which is the default) includes the fake prefixes.

Or am I misunderstanding things?

OJFord:
> From you above example [code] ... looks to me like they (a/ and b/) were not stripped

That's the `--no-pager` output though, see below it for with dsf, and no `a/`, `b/`.

> Just so I'm being 100% clear... when you set `git config diff.noprefix true`

Yes, `noprefix=true` should be _not have_ `a/`, `b/`; `noprefix=false` should _have_ `a/`, `b/` - and dsf should change the default (`noprefix` unset) behaviour to match as if `noprefix=true` specified.

I don't use Perl, but my reading of this PR was that ^that's what it did do - I just can't make it work in practice.

scottchiefbaker:
At no point does d-s-f EVER show the prefix. We make the output as human readable as possible.

The issue this pull request solves is when a file is in a REAL directory `a/` or `b/` and d-s-f was stripping that thinking that it was a prefix dir. This pull request does NOT strip the REAL directory `a/`.

You can test this by setting noprefix to true and then creating "a/foobat.txt".  Commit that, and then changing that file. Notice that d-s-f reports the file as "foobar.txt" **not** "b/foobar.txt".

OJFord:
> At no point does d-s-f EVER show the prefix. We make the output as human readable as possible.

Oh. Then I misunderstood the purpose of this PR.

IMO respecting a user's `diff.noprefix` means _not_ removing `a/` and `b/` if they explicitly set `noprefix=false`. My proposal in #205 was that dsf just change the behaviour when `noprefix` has not been explicitly set by the user, treating it as `true` instead of `false` as git does, essentially.

If that's the way you want to go, then yep, this works for me. But I think dsf making things "as human readable as possible" should not take precedence over a user's explicit configuration.

Issue #223, opened 2017-03-03
Diff-so-fancy needs usage output


scottchiefbaker:
@stevemao wrote

> make diff-so-fancy as a standalone cli command sounds great. Would like to see at least some information printed when running it instead of justing hanging there.

I actually just figured out how to detect that status of (or lack of) STDIN in Perl, so this is very doable. What would like d-s-f to output if it's run as raw from the command line?

stevemao:
I was just talking about `diff-so-fancy --files file1 file2` raised in https://github.com/so-fancy/diff-so-fancy/issues/220
If you just run `diff-so-fancy` it should print help? Similar to what https://github.com/so-fancy/diff-so-fancy/pull/218 is printing.

scottchiefbaker:
Good point... I clarified the name of the issue. I just need standard usage for calling `diff-so-fancy` without any `STDIN`.

Do we just want to output a chunk of the readme?

OJFord:
Somewhat related to help output, it'd be nice to have `-v/--version` work.

scottchiefbaker:
I could probably do both at the same time. The only problem with `--version` is that we have to maintain the version number inside of `d-s-f`.

scottchiefbaker:
I mocked up a sample that handles both `--version` and calling d-s-f without any arguments.

scottchiefbaker:
Closing via #242

PR #224, opened 2017-03-06
Initial checkin for --version and CLI usage

scottchiefbaker:
Should address #223


scottchiefbaker:
```
:perl libexec/diff-so-fancy.pl --version
Version      : 0.12.0
```

and

```
:perl libexec/diff-so-fancy.pl
Usage:

  # One off fanciness
  git diff --color | diff-so-fancy

  # Configure git to use d-s-f for all diff operations

  # Create a git alias to run d-s-f on demand
  git config --global alias.dsf '!f() { [ -z "$GIT_PREFIX" ] || cd "$GIT_PREFIX" && git diff --color "$@" | diff-so-fancy  | less --tabs=4 -RFX; }; f'
```

stevemao:
Great work! You can now remove anything related to Bash such shellcheck in the codebase. 😄 
cc @paulirish 
I think we can release v1 after this. Initially I was waiting on #35, but it seems that it'll take a while. The project should already be stable 😃

scottchiefbaker:
I was waiting 100% for the CI tests to run before I said anything. The local `bats` tests run fine, so I'm sure this will as well.

🥇 🥇 🥇 🥇 🥇 🥇 🥇 🥇 🥇 🥇 🥇 🥇 🥇 🥇 🥇 

As you pointed out this removes the d-s-f bash script, because the **entire** thing is Perl now. This should make it ever so slightly faster as well. The only functional change is that we now depend on `which` to find the location of `diff-highlight`. Hopefully that won't have any negative effects (especially on Windows), but it's hard to say.

@stevemao @OJFord @paulirish can you test this version on your respective OSs? Specfically OSX to make sure things work.

@arichiardi will you update #217 to reflect these new changes. It shouldn't be anything major, just a couple of paths that changed.

arichiardi:
So basically I need to point to the new `.pl` file instead, which will be run instead of the `.sh`. Correct?

scottchiefbaker:
@arichiardi yes... there should be no `/libexec` folder anymore. The only file you need to worry about is the `diff-so-fancy` file in the root of the repo. It's a Perl script without the extension.

Well I guess if you want to download `diff-highlight` to that's a good idea.

arichiardi:
Ok thanks, I will try over the next couple of days.

scottchiefbaker:
@peff now that d-s-f is entirely Perl, I'd like to move the `diff-highlight` coloring inside of d-s-f as well. Did you write `diff-highlight`? I see you were the last commiter, but that doesn't me you wrote it. I'd like to work on an implementation of d-s-f with all that logic inline (i.e. not a separate script). Do you know who I'd talk to about that?

peff:
@scottchiefbaker Yes, I wrote it. There were a few minor fixes from other people, but nothing really substantial[1].  So as far as licensing goes, I'm happy for you to take whatever bits of it you'd like.

[1] `git blame` can tell you, or `git log --invert-grep --author=peff -p contrib/diff-highlight`

diff-so-fancy
```diff
@@ -1,33 +1,565 @@
...
+	if ($_[0]) { return $ret->{$_[0]}; }
+
+	return $ret;
+}
+
+# Output the command line usage for d-s-f
+sub usage {
+	my $out = "Usage:
+
+  # One off fanciness
```
paulirish: should we put this in a little text file? Might be easier to maintain

diff-so-fancy
```diff
@@ -1,33 +1,565 @@
...
+  git diff --color | diff-so-fancy
+
+  # Configure git to use d-s-f for all diff operations
+  git config --global core.pager \"diff-so-fancy | less --tabs=4 -RFX\"
+
```
paulirish: I'd also add these:

```sh
# If core.pager config is set, this can opt-out of using d-s-f
git --no-pager <command>
```

and

```sh

# Print version metadata
diff-so-fancy --version
```

scottchiefbaker:
@paulirish I understand the diff is pretty messy, but that's because we removed the old bash script and moved the Perl one in it's place. I don't know of a clean git way to do that.

For the purposes of blame I wrote the **entire** thing, so it all falls back on me. Worst case scenario a line might not map to a specific commit, but I think git handles moves pretty well so we should be ok.

paulirish:
> but that's because we removed the old bash script and moved the Perl one in it's place. 

ah. i suppose in this commit https://github.com/so-fancy/diff-so-fancy/pull/224/commits/f1ae25111016f6a4a213b5214eb0a12f40413955 you could have `git mv`'d the .`pl` into the new filename and i think it'd end up with minimal diff.

scottchiefbaker:
I did `git mv` the file. I'm 99% sure it will look fine when it's all done, it's just the diff for this changeset that looks gnarly.

peff:
If you add `-B` to your diff-options, Git will find the rename (and show the additional modifications on top of the renamed original). The issue is that Git's rename detection by default only looks for matches between files that went away and files that appeared. In your case, `diff-so-fancy` didn't appear; it was already there with other content. `-B` turns on "break detection", which notices that the file was basically rewritten, and turns it into a delete/add pair, which is then fed to the rename detection.

Likewise, if you use the `-C` option to `git blame`, it will detect the movement of lines between files.

scottchiefbaker:
Any reason not to land this?

scottchiefbaker:
I've been testing this for the last week and it's been working great. Unless there are objections I'm going to land this tomorrow.

Paging: @paulirish @OJFord @stevemao

paulirish:
Sgtm

stevemao:
Sorry for the delay. Looks like `git log` is not working with this anymore?

scottchiefbaker:
Do we support using d-s-f with `git log`? I don't think I've ever used it in that fashion. You must have some log alias that uses d-s-f, because vanilla `git log` works just fine for me.

stevemao:
We had an issue on that before, and eventually we fixed it. you could print diffs with `git log --full-diff -p .`

scottchiefbaker:
Well the gurus on #perl set me straight on how to properly feed data to `diff-hightlight` and now `git log --full-diff -p .` doesn't block. Can you try now?

stevemao:
Looks it's working now 😄 Cheers!

PR #227, opened 2017-03-26
update manual install details

techgaun:
I guess `libexec/diff-so-fancy.pl` was an old structure and does not seem to relevant.


scottchiefbaker:
You are correct. We're still in the process of updating all the docs. That change only happened within the last two weeks, and no official release has taken place yet. Thanks for the PR.

techgaun:
ahh okay that makes sense. cool, thanks

scottchiefbaker:
I would love a fresh set of eyes on our documentation (README.md) to see if it's still relevant for a new user. We'd welcome any other suggestions you have.

PR #228, opened 2017-03-29
Remove reference to libexec in readme.md

marema31:
In the manual installation there is reference to a libexec directory that is not available on the repository.


scottchiefbaker:
Good catch. We just landed some file move changes and I haven't had a chance to update the documentation yet.

PR #230, opened 2017-05-08
Add Dockerfile

volker-fr:
This pull request should add docker support to diff-so-fancy.

If this PR gets approved and merged, the next step should be to create an official docker repository where a new `diff-so-fancy` container gets automatically build on each new push to the master branch.

Feel free to update the maintainer in the Dockerfile.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+# USAGE:
+#   git config --global core.pager "docker run --rm -i diff-so-fancy| less --tabs=4 -RFX"
+
+FROM alpine
+MAINTAINER Volker <lists.volker@gmail.com>
+
+RUN apk --no-cache add perl ncurses git
+
+ADD third_party/diff-highlight/diff-highlight /usr/local/bin/diff-highlight
+ADD diff-so-fancy /usr/local/bin/diff-so-fancy
+
+ENV TERM=vt100
+
+ENTRYPOINT [ "/usr/local/bin/diff-so-fancy" ]
--- a/readme.md
+++ b/readme.md
@@ -73,6 +73,19 @@ If you want, you can choose to install manually:
 
 Note: The `diff-highlight` dependency is an [official git-contrib script](https://github.com/git/git/tree/master/contrib/diff-highlight), duplicated here for convenience. If you prefer less fancy in your diff, you also use diff-highlight [on it's own](https://news.ycombinator.com/item?id=11068436).
 
+## Docker install
+You can run `diff-so-fancy` in a docker container. Downside of this approach is that `diff-so-fancy` doesn't have any direct access on the git repository to load its configuration. Benefit is that there will be hopefully a public docker container in the future which only requires you to run one simple `git config` command to set up `diff-so-fancy`.
+
+```shell
+# Clone this repository
+git clone https://github.com/so-fancy/diff-so-fancy.git
+# Build the docker container
+cd diff-so-fancy
+docker build -t diff-so-fancy .
+# Configure git to use diff-so-fancy
+git config --global core.pager "docker run --rm -i diff-so-fancy| less --tabs=4 -RFX"
+```
+
 ## options
 
 ### markEmptyLines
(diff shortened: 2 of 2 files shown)


paulirish:
since this utility is rather small, i'm surprised there'd be demand for running it via docker.

scottchiefbaker:
I agree with @paulirish ... what's the real world use case for running d-s-f in Docker?

volker-fr:
I try to run as much as possible in Docker on my workstations. Makes reinstalling systems as well as updating them much quicker and easier, especially if its across systems or a package is not available for a distribution. E.g. I haven't seen it for Debian.

Hence I created it for my own usage and created a PR in case someone else is interested and it doesn't interfere with the existing code.

I also saw that [someone else](https://hub.docker.com/r/thunderk/diff-so-fancy/) seemed to have considered but there isn't an existing image yet.

Dockerfile
```diff
@@ -0,0 +1,14 @@
+# USAGE:
+#   git config --global core.pager "docker run --rm -i diff-so-fancy| less --tabs=4 -RFX"
```
elamperti: Is it necessary to have this comment here, considering that people using this Dockerfile can check the readme?

readme.md
```diff
@@ -73,6 +73,19 @@ If you want, you can choose to install manually:
...
 
+## Docker install
+You can run `diff-so-fancy` in a docker container. Downside of this approach is that `diff-so-fancy` doesn't have any direct access on the git repository to load its configuration. Benefit is that there will be hopefully a public docker container in the future which only requires you to run one simple `git config` command to set up `diff-so-fancy`.
+
+```shell
+# Clone this repository
+git clone https://github.com/so-fancy/diff-so-fancy.git
+# Build the docker container
+cd diff-so-fancy
+docker build -t diff-so-fancy .
```
elamperti: If the repo is added as a docker hub image these instructions could be replaced by a simple `docker pull` of that image. In my case, I'd think about using the docker image only if it spared me from having to clone this repo locally in the first place.

OJFord:
> I try to run as much as possible in Docker on my workstations.

I'm really curious how this even works? Git also runs in a (separate) container?

> Makes reinstalling systems as well as updating them much quicker and easier

It seems like you must pay for it in setup though? Creating the images, because nobody else runs dsf (for example) in a docker container, aliasing each program to a docker command that will run its containerised version, etc.?

It sort of appeals to me, which is why I'm asking about it, but unless I'm missing something I just can't see that it's practical with Docker.

readme.md
volker-fr: Correct. That's the plan. I actually use my own docker hub repo for that but it's not an official one. That would be the next step.

Dockerfile
volker-fr: I always write the usage in top just in case. It can be hidden in the README or a lot of public docker repositories don't come with a readme that explains usage.

volker-fr:
> I try to run as much as possible in Docker on my workstations.
    I'm really curious how this even works? Git also runs in a (separate) container?

No. There are shortcomings where I can't use everything in a container without a lot of pain points. git & the editor of choice are some of them.

    > Makes reinstalling systems as well as updating them much quicker and easier
    It seems like you must pay for it in setup though? Creating the images, because
    nobody else runs dsf (for example) in a docker container, aliasing each program
    to a docker command that will run its containerised version, etc.?

You have to set it up once, but afterwards no more. You can either do that by auto-installing it via curl or similar or with docker. I prefer docker for various different reasons. And if you have your dot files somewhere you just clone from and then you run a script to link the dotfiles to the place -- like you could run an installer script that uses curl to download it.

A container should come with everything required. I don't have to deal with specific versions of $language-of-choices, especially their dependencies if I have it in a container. And for d-s-f  I am happy to just create it myself and host it myself. I use it now for a few weeks and am happy with it. I don't have to worry if the next base system removed now perl, comes with ncurses or not. I don't have to worry about dependencies or try to find yet-again out how to install it.

     It sort of appeals to me, which is why I'm asking about it, but unless I'm missing something I just can't see that it's practical with Docker.

Overhead of docker is small, benefits of not dealing with software & its dependencies is appealing to me, and even if there is a bug the containers are usually as small as possible and have usually only the necessary files, if at all, linked into the container. There are shortcomings with having everything in docker, but there are also benefits.

It seems like there is no interest in merging this which is fine with me. I just host it myself and use it myself. Feel free to close this ticket.

scottchiefbaker:
We didn't come to a consensus on this, and I don't know enough about docker to speak intelligently. Closing, but willing to reopen if it's an issue for other people.

PR #231, opened 2017-05-10
Fix curl flags in update-deps.sh

oranja:
After `-o` must follow the output path.
The `-L` should either come before `-o` or after the path.


scottchiefbaker:
Good catch

PR #232, opened 2017-05-29
re-enable shellcheck on CI

paulirish:
it was failing on circleCI: https://circleci.com/gh/so-fancy/diff-so-fancy/404
and just not enabled on travis

So i've flipped it back on in the config, included report-bug.sh and fixed the errors there.

Now we're green on all our CI. \o/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,3 +33,4 @@ install:
 script:
   - ./bats/bin/bats test/*.bats
   - perlcritic -1 -q --theme freenode diff-so-fancy
+  - shellcheck *.sh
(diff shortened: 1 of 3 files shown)


scottchiefbaker:
I'm pretty sure I disabled this... we don't have any .sh files anymore, do we need to try and spellcheck them?

stevemao:
@scottchiefbaker I think we can just remove `shellcheck`

paulirish:
well it's there for report-bug.sh now. and I did fix a number of style issues that seem worthwhile.

i think having it present to increase confidence in our bug reporting flow is worth it.

scottchiefbaker:
@paulirish I totally forgot about the `report-bug.sh` good call. I'll merge this.

PR #234, opened 2017-06-06
Add Perl 5.26 compatibility

### LINKED ISSUE #233: Does not work with Perl 5.26
I just updated my `Arch Linux` and with it came `Perl 5.26`. This is the error I receive with it:

```
Unescaped left brace in regex is deprecated here (and will be fatal in Perl 5.30), passed through in regex; marked by <-- HERE in m/^((?^:(\e\[([0-9]{1,3}(;[0-9]{1,3}){0,3})[mK])?))[ +-]{ <-- HERE -1}/ at /usr/lib/node_modules/diff-so-fancy/libexec/diff-so-fancy.pl line 364, <> line 6.
```

According to a bunch of places on the Internet, unescaped braces are a syntax error from `5.26` on. This is rather urgent, could you fix it, please, and release a new version? :)

boris-petrov:
Fixes #233

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -133,7 +133,9 @@ while (my $line = <$input>) {
 		# The number of colums to remove (1 or 2) is based on how many commas in the hunk header
 		$columns_to_remove   = (char_count(",",$hunk_header)) - 1;
 		# On single line removes there is NO comma in the hunk so we force one
-		$columns_to_remove ||= 1;
+		if ($columns_to_remove <= 0) {
+			$columns_to_remove = 1;
+		}
 
 		if ($1) {
 			print $1; # Print out whatever color we're using


scottchiefbaker:
I think you're on the right track here, but I don't think this is the correct fix. On that line we're saying:

Replace chars after a space/plus/minus X times.

Where X $columns_to_remove. I think maybe it doesn't like the nested { }.

scottchiefbaker:
I'm guessing that `$columns_to_remove` is getting set to "" or undef and it's causing this error. Can you try changing: 

    my $columns_to_remove = shift();

to

    my $columns_to_remove = int(shift());

and testing? That should 100% ensure that it's an integer. I don't have 5.26 yet to test with.

scottchiefbaker:
Actually that error exists in 5.22 and 5.24 (which we test with), it's just marked as "will be fatal in 5.30" in the newest version of Perl. In other words this is not a Perl 5.26 bug, it's a logic bug.

Can you provide the raw diff that you're generating that's causing this error?

https://github.com/so-fancy/diff-so-fancy#reporting-bugs

benoit-pierre:
The issue occurs when running the testsuite:
```
 ✗ Handle unicode characters in diff output
   (in test file test/diff-so-fancy.bats, line 80)
     `output=$( load_fixture "unicode" | $diff_so_fancy )' failed with status 25
   Unescaped left brace in regex is deprecated here (and will be fatal in Perl 5.30), passed through in regex; marked by <-- HERE in m/^((?^:(\e\[([0-9]{1,3}(;[0-9]{1,3}){0,3})[mK])?))[ +-]{ <-- HERE -1}/ at /home/bpierre/progs/src/diff-so-fancy/test/../diff-so-fancy line 388, <$cmd_output> line 6.
```
The issue is apparently that `$columns_to_remove` is negative and so the `{-1}` in the regexp triggers the warning.

benoit-pierre:
Which is because `test/fixtures/unicode.diff` does not have a comma in its hunk:
```diff
diff --git a/unicodes b/unicodes
index 223f57d..1c2066d 100644
--- a/unicodes
+++ b/unicodes
@@ -1 +1 @@
-aao
+åäöç
```

I guess a proper fix would be:

```diff
diff --git i/diff-so-fancy w/diff-so-fancy
index 7c1cec2..5a6af6d 100755
--- i/diff-so-fancy
+++ w/diff-so-fancy
@@ -133,7 +133,9 @@ while (my $line = <$input>) {
                # The number of colums to remove (1 or 2) is based on how many commas in the hunk header
                $columns_to_remove   = (char_count(",",$hunk_header)) - 1;
                # On single line removes there is NO comma in the hunk so we force one
-               $columns_to_remove ||= 1;
+               if ($columns_to_remove <= 0) {
+                       $columns_to_remove = 1;
+               }

                if ($1) {
                        print $1; # Print out whatever color we're using
```

scottchiefbaker:
`$columns_to_remove` is getting set to -1 somehow. This is causing the regexp to break like @benoit-pierre pointed out. The real question is WHY is it getting set to -1? We calculate `$columns_to_remove` by looking at the hunk header and counting the number of [commas](https://github.com/so-fancy/diff-so-fancy/blob/master/diff-so-fancy#L134). This should determine if we're in a two-way or three-way merge. If it's ending up as -1 then it must be parsing the hunk header incorrectly.

Can you post the raw diff so can figure out why d-s-f is calculating incorrectly?

boris-petrov:
I tried the easiest fix which seemed to work but now I see that actually it doesn't work perfectly fine - there are `+` and `-` signs in the beginning of each diffed line. So you're right, this is not the correct fix.

I'm not sure I follow the conversation as I don't know the code, but let me know if I can do something to help - change the PR or something. :)

@scottchiefbaker - I tried changing the `my $columns_to_remove = shift();` line but the result was the same - `+` and `-` in the beginning of the line.

The interesting part is that this error happens only on specific diffs (I guess which contain `{`?). This diff for example hits the issue (a JSON file):

```diff
-[{"a": "b"}]
+[{"a": "c"}]
```

scottchiefbaker:
@boris-petrov can you post the entire diff in question? I need to see the hunk header (the part in the @@@) and some of the code to figure this out.

boris-petrov:
@scottchiefbaker - if you are talking about the diff I pasted at the bottom of my comment, this is the full diff. Just create a new file with `[{"a": "b"}]` in it, add it to a git repo, then change `b` to `c` and you will see the issue.

benoit-pierre:
What about the example in the test suite itself: `test/fixtures/unicode.diff`?

benoit-pierre:
Which is valid:

```bash
> mkdir {a,b}
> echo aao > a/unicodes
> echo åäöç > b/unicodes
> echo åäöç > b/unicodes
> diff -u {a,b}/unicodes
--- a/unicodes  2017-06-07 16:12:04.552188801 +0200
+++ b/unicodes  2017-06-07 16:12:06.632188846 +0200
@@ -1 +1 @@
-aao
+åäöç
> diff -u {a,b}/unicodes | ./diff-so-fancy
modified: unicodes
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
@ unicodes:1 @
Unescaped left brace in regex is deprecated here (and will be fatal in Perl 5.30), passed through in regex; marked by <-- HERE in m/^((?^:(\e\[([0-9]{1,3}(;[0-9]{1,3}){0,3})[mK])?))[ +-]{ <-- HERE -1}/ at ./diff-so-fancy line 388, <$cmd_output> line 4.
```

Again, the patch in my comment above fix the issue.

boris-petrov:
I can confirm that @benoit-pierre's patch fixes the issue. I've updated the pull-request.

scottchiefbaker:
This is 100% the fix, you guys are correct. This appears to not **always** trigger a warning until Perl 5.22 (5.20 and previous are silent). I'm curious why our test suite doesn't catch this:

Perl 5.22 output: https://travis-ci.org/so-fancy/diff-so-fancy/jobs/240392630

Any ideas? If we can figure out why the automated testing isn't working I'll gladly land this PR.

I'll add Perl 5.26 Travis testing, as well as bleed just to be safe.

benoit-pierre:
But 5.22 does not output a warning, I've tested locally by running the Perl 5.22 binary you use on Travis, and there's no warning.

I think a specific test should be added:

```diff
 test/diff-so-fancy.bats          | 8 ++++++++
 test/fixtures/hunk_no_comma.diff | 7 +++++++
 2 files changed, 15 insertions(+)

diff --git c/test/diff-so-fancy.bats i/test/diff-so-fancy.bats
index 7c1b675..2c2c3a0 100644
--- c/test/diff-so-fancy.bats
+++ i/test/diff-so-fancy.bats
@@ -87,3 +87,11 @@ if begin[m"
   # Make sure the output contains SOME of the english text (i.e. it doesn't barf on the whole line)
   assert_output --partial "saw he conqu"
 }
+
+@test "Correctly handle hunk definition with no comma" {
+  output=$( load_fixture "hunk_no_comma" | $diff_so_fancy )
+  # On single line removes there is NO comma in the hunk so we force one,
+  # make sure the first column is still correctly stripped.
+  run printf "%s" "$output"
+  assert_line --index 5 "after"
+}
diff --git c/test/fixtures/hunk_no_comma.diff i/test/fixtures/hunk_no_comma.diff
new file mode 100644
index 0000000..4f4445f
--- /dev/null
+++ i/test/fixtures/hunk_no_comma.diff
@@ -0,0 +1,7 @@
+diff --git i/file w/file
+index 90be1f3..294186e 100644
+--- i/file
++++ w/file
+@@ -1 +1 @@
+-before
++after
```

benoit-pierre:
Or maybe some of the tests in `test/diff-so-fancy.bats` should be reworked to be less lenient, maybe by stripping color information so exact matches can be used with `assert_line` and the like?

benoit-pierre:
To be clear:
```perl
"test" =~ /test{-1}/;   # warning with 5.22 and 5.26
"test" =~ /(test){-1}/; # no warning with 5.22
```

scottchiefbaker:
@benoit-pierre I just tested on Perl 5.24 on Fedora 25 and I don't get an error either. Must be very version/situation specific. I definitely think we should add a test for this. I'll land this fix now, 

@benoit-pierre can you open a separate PR for the test you have above?

benoit-pierre:
Done: #235.

PR #235, opened 2017-06-07
add a test for #233

benoit-pierre:
Check the first column is correctly stripped when a hunk has no comma.

### DIFF
--- a/test/diff-so-fancy.bats
+++ b/test/diff-so-fancy.bats
@@ -87,3 +87,11 @@ if begin[m"
   # Make sure the output contains SOME of the english text (i.e. it doesn't barf on the whole line)
   assert_output --partial "saw he conqu"
 }
+
+@test "Correctly handle hunk definition with no comma" {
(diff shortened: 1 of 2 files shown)


scottchiefbaker:
Just so I understand... we're checking that line 5 says "after"? The logic here being, if there was that regexp error it would drop the after down a couple lines and this test would fail?

benoit-pierre:
Here is the result of running the test without the fix:
```
 ✗ Correctly handle hunk definition with no comma
   (from function `assert_line' in file test/test_helper/bats-assert/src/assert.bash, line 477,
    in test file test/diff-so-fancy.bats, line 96)
     `assert_line --index 5 "after"' failed

   -- line differs --
   index    : 5
   expected : after
   actual   : +after
   --
```

benoit-pierre:
I could check line 4 too, for completeness.

scottchiefbaker:
Your test for line 5 should be enough. Thanks for doing this!

PR #240, opened 2017-06-26
Wrap diff_highlight in quotes to avoid err on windows

### LINKED ISSUE #239: [Windows] Can't exec "/c/Program": No such file or directory
Installed using npm, running using [cmder](https://github.com/cmderdev/cmder) on Windows 8.

Error:
```
λ git diff --color | diff-so-fancy
Can't exec "/c/Program": No such file or directory at C:\Program Files\nodejs\node_modules\diff-so-fancy\diff-so-fancy line 568.
```

sudo-suhas:
Not sure if this is the best solution but fixes the issue(#239) on Windows 8.

Ran tests using mintty on cmder. 1 test fails:
```
not ok 9 header format uses a native line-drawing character
# (from function `assert_line' in file test/test_helper/bats-assert/src/assert.bash, line 468,
#  in test file test/diff-so-fancy.bats, line 51)
#   `assert_line --index 0 --partial "─────"' failed
#
# -- line does not contain substring --
# index     : 0
# substring : ─────
# line      : --------------------------------------------------------------------------------
# --
#
```
But **all** tests fail on the master branch.

Also, not sure if this can cause any issue on other platforms.

Closes #239

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -565,7 +565,7 @@ sub filter_stdin_through_diff_highlight {
 	my $dh  = find_diff_highlight();
 
 	# Pipe the STDIN of this to diff-highlight
-	my $pid = open(my $cmd_output, "-|", $dh);
+	my $pid = open(my $cmd_output, "-|", "\"$dh\"");
 
 	return $cmd_output;
 }


scottchiefbaker:
Hmmmmmmmm... the only test that failed as a build on OS X, which may have been a fluke.

I think the fix may be better as: 

    my $pid = open(my $cmd_output, "-|", "\"$dh\"");

Can you update your patch and resubmit? With this change, the tests don't fail on my Linux box.

sudo-suhas:
Done. I removed the previous commit. Hope that's okay. Less noise.

scottchiefbaker:
I don't think we've ever run (or even tried) to run the tests on Windows. We're sorely lacking a Windows CLI expert. All the main devs use either Linux or Mac OS X.

sudo-suhas:
Ha ha. I can understand why. I wish I wasn't on windows either!

scottchiefbaker:
Tests run for me with this modified patch. It shouldn't be needed on Linux or OS X, but if it fixes things for Windows I'll land it. @sudo-suhas does this fix the issue for you?

sudo-suhas:
It does. Can't use without it. Would be okay if I installed to path without space I guess.

scottchiefbaker:
Ok I'll land this for now since it fixes Windows, and doesn't affect Linux / OS X

stevemao:
Do we have any test missing on Windows? CI didn't fail before this PR.

scottchiefbaker:
All tests are passing just fine. The first "failure" was a fluke on Travis. If you look at the results you can see it was bogus.

sudo-suhas:
@stevemao All the current tests were failing on my local. I think it is less to do with the tests themselves and more to do with the specific setup. Right now, all except 1 test pass.

PR #256, opened 2017-07-31
fix #254: Displays wrong commands on initial screen in usage

### LINKED ISSUE #254: Displays wrong commands on initial screen in usage
See the second and third commands under "Usage". It should be `diff-so-fancy` instead of `git diff`.

```
Version      : 1.1.1

Usage:

git diff --color | diff-so-fancy # Use d-s-f on one diff
git diff --colors                # View the commands to set the recommended colors
git diff --set-defaults          # Configure git-diff to use diff-so-fancy and suggested colors

# Configure git to use d-s-f for *all* diff operations
```

rksys:
fix issue #254

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -552,8 +552,8 @@ sub usage {
 	$out .= "Usage:
 
 git diff --color | diff-so-fancy # Use d-s-f on one diff
-git diff --colors                # View the commands to set the recommended colors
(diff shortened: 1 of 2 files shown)


rksys:
@scottchiefbaker on next branch now, with comment # aligned ;-)

scottchiefbaker:
Perfect!

PR #257, opened 2017-07-31
submit PR against the next branch

stevemao:


scottchiefbaker:
Did you create an actual `next-branch` branch? :)

I checked 7d6549e29abcfc5408bd3a2c037ab137e3252fe2 in to `next` earlier today, but your wording/link may be better. What do you think?

stevemao:
😄 Up to you

scottchiefbaker:
Agreed... why don't you modify what I already have and add your link. It'll be the best of both worlds.

PR #260, opened 2017-08-14
update-deps: Fix DiffHighlight's URL and refactor

oranja:
Seems like I'm the only one who tries to use `update-deps.sh`, huh?
This time there were two different problems:
1. `diff-highlight` was made a `.pm` so the old URL doesn't point to a valid endpoint.
2. `diff-highlight`/`DiffHighlight` is no longer expected to be found locally in `third-party`, but in `lib`. I guess that's some `perl` or at least `fatpack` convention for where local modules should be placed.

A word of warning, though: Unlike this repo's version of [`DiffHighlight`](/so-fancy/diff-so-fancy/blob/master/lib/DiffHighlight.pm) , the current git-contrib version of [`DiffHighlight`](/git/git/blob/master/contrib/diff-highlight/DiffHighlight.pm) doesn't `use Encode`.
This means that after an update & rebuild (fatpack), `diff-so-fancy` will fail ([here](/so-fancy/diff-so-fancy/blob/master/diff-so-fancy#L474)) unless it has its own `use Encode`.

_Source: It bit me._ :/

### DIFF
--- a/update-deps.sh
+++ b/update-deps.sh
@@ -1,4 +1,11 @@
 #!/bin/bash
 
-curl -Lo "third_party/diff-highlight/diff-highlight" "https://github.com/git/git/raw/master/contrib/diff-highlight/diff-highlight"
-curl -Lo "third_party/diff-highlight/README" "https://github.com/git/git/raw/master/contrib/diff-highlight/README"
+DIFFHIGHLIGHT_RAW_URL_BASE="https://raw.githubusercontent.com/git/git/master/contrib/diff-highlight"
+DIFFHIGHLIGHT_FILES=( "DiffHighlight.pm" "README" )
+
+for file in "${DIFFHIGHLIGHT_FILES[@]}";
+do
+  url="$DIFFHIGHLIGHT_RAW_URL_BASE/$file"
+  echo "$url"
+  curl -#Lo "lib/$file" "$url"
+done


scottchiefbaker:
I appreciate you giving `update-deps.sh` some love. It doesn't get a lot of attention.

1. You're absolutely correct
2. Previously `diff-highlight` was a stand alone script that we shelled out to run. It's been made in to a module for performance improvements. This is not a `fatpack` thing, but a general Perl thing.

`fatpack` is now used to "bundle" the script and the `DiffHighlight.pm` dependency in to one script for simplicity. Can you elaborate on the `git-contrib` piece, I don't totally follow what you're trying to convey.

scottchiefbaker:
The `use Encode` thing is weird... it's included in `DiffHighlight.pm` so the main script should be able to access those functions also. I did update the main script to explicitly call it anyway.

oranja:
You already fixed it, but I'll try to be clearer.

In this repo (diff-so-fancy), **DiffHighlight.pm** has a `use Encode;` line that doesn't exist in **DiffHighlight.pm** from the official `Git` repository (in **contrib/diff-highlight**).

The base `diff-so-fancy` script (before `fatpack`-ing), calls `Encode::encode()` but doesn't have a `use Encode;` line.

If I don't run `update-deps.sh`, the final fatpacked version has a `use Encode;` line, as you said, and everything is OK,
but if I do run `update-deps.sh`, the final fatpacked won't have this line anymore and it will exit with an error.

Your fix (#260) prevents the error.

scottchiefbaker:
Oh! Upstream `DiffHighlight.pm` makes more sense. I've diverged slightly from upstream. I submitted some changes to @peff but he's been pretty busy. 

I'll gladly merge this, if you change it to be against `next`. `master` is stable and released, anything other than emergency bugfixes should be directed against `next`.

oranja:
Rebased upon `next`.

Upstream. I should have just said upstream :nerd_face:

PR #267, opened 2017-09-29
[Windows] Fix incorrect directory listing for UNIX emulated perl on Windows

### LINKED ISSUE #265: Can't locate DiffHighlight.pm in @INC (Windows 10)
Hi,

I'm having troubles installing this plugin on win10.

I'm using Git for Windows (portable version) and PowerShell.
Copied diff-so-fancy folder to PortableGit folder and executed from PowerShell following command: `git config --global core.pager "'C:\Portable\PortableGit\diff-so-fancy\diff-so-fancy' | less --tabs=4 -RFX"`

Can you please add a clearer instruction for installation on Win10.

Thank you.

stanislavromanov:
This will close #265 

Using UNIX emulated perl on Windows (which many people use) will cause `diff-so-fancy` not work because the path resolution is wrong.

This is the output we get from `dirname(abs_path($0)) . "/lib"`:
```
/c/Users/stani/Work/Personal/diff-so-fancy/C:/Users/stani/AppData/Roaming/npm/node_modules/diff-so-fancy/lib
```

Clearly this path is invalid therefore the module `DiffHighlight` is not found. By using `dirname(File::Spec->catdir($0))` we get the correct path.

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -4,10 +4,10 @@ my $VERSION = "1.2.0";
 
 #################################################################################
 
-use Cwd qw(abs_path);                   # For realpath()
-use File::Basename;                     # for dirname
-use Encode;                             # For handling UTF8 stuff
-use lib dirname(abs_path($0)) . "/lib"; # Add the local lib/ to @INC
+use File::Spec;                                     # For catdir
+use File::Basename;                                 # For dirname
+use Encode;                                         # For handling UTF8 stuff
+use lib dirname(File::Spec->catdir($0)) . "/lib";   # Add the local lib/ to @INC
 use DiffHighlight;
 
 use strict;


scottchiefbaker:
This looks good. I didn't know about `File::Spec`, it looks like a good solution. If I check the diff the alignment of the comments go skewed in this patch. If you can clean that up I'll gladly merge this.

stanislavromanov:
@scottchiefbaker I made comments use the same alignment for all the `use` cases. I'm not use what is wrong with it. It looks like this in the end:

```perl
use File::Spec;                                     # For catdir
use File::Basename;                                 # For dirname
use lib dirname(File::Spec->catdir($0)) . "/lib";   # Add the local lib/ to @INC
use DiffHighlight;
```

Is it not OK with you?

scottchiefbaker:
If I click on "files changed" in this Pull Request to view the actual diff the alignment is all out of whack. Maybe that's just a github thing?

scottchiefbaker:
@stanislavromanov I just realized that this breaks symlinks. For example on my server I have symlink in `~/bin/` that points to `~/github/diff-so-fancy`. This patch adds `~/bin/lib` to the path instead of the real path `~/github/diff-so-fancy/lib/`

This used to work when we did `abspath()`... I'm not sure if we can have it both ways.

stanislavromanov:
Okey, I will continue using this on my machine because it fixes the bug for me. But at least you know *where* is the problem so maybe we can check if there is some other way.

I know for a fact that what else could work is just taking the `$0` and replacing `\` signs with `/` (because windows) but that didn't felt right :)

scottchiefbaker:
@stanislavromanov/@jedi-tester can you test by changing the header to look like so:

```
use Cwd qw(abs_path);                                       # For realpath()
use lib dirname(abs_path(File::Spec->catdir($0))) . "/lib"; # Add the local lib/ to @INC
```

This works for me on Linux, I'm curious if it works on Windows too.

scottchiefbaker:
Pinging @stanislavromanov @jedi-tester 

Can you test this last change I proposed on Windows. It should fix both symlinks and Windows URLs.

stanislavromanov:
@scottchiefbaker Just tested, your code works.

scottchiefbaker:
Ok great... I just landed that code too.

PR #275, opened 2017-11-03
Add offset for pager line numbers

alok:
Feel free to ignore, just writing this up in case you decide you do want to do this.


scottchiefbaker:
I'm reluctant to merge this. The point of the divided is that it spans the full width of the screen. Visually it's more appealing that way.

The use case of compensating for line numbers is pretty minimal.

PR #276, opened 2017-11-09
Sparrow tests for diff-so-fancy tool

melezhik:
Hi guys! I've noticed you use bats for tests. Are you satisfied with it? Alternatively  this is how tests might be achieved by Sparrow framework. Features of possible interest are:

* checking against Perl5 regexps
* checking in sequences of lines (LineA,LineB,LineC so on  - see story.check file )
* Under the hood it's still Bash scripts + check files, so it's pretty simple 

However if you're pretty satisfied with Bats - that's ok, just close this request.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,8 +31,10 @@ before_install:
 install:
   - cpanm --quiet --notest --local-lib=~/perl5 local::lib
   - eval $(perl -I ~/perl5/lib/perl5/ -Mlocal::lib)
-  - cpanm --quiet --notest Test::Perl::Critic Perl::Critic::Freenode
+  - cpanm --quiet --notest Test::Perl::Critic Perl::Critic::Freenode Sparrow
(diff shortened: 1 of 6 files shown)


scottchiefbaker:
I'm really not sure what this offers us over what we already have. I'm pretty happy with `bats`.

melezhik:
Hi! Probably you don't need but I coils find regexp checks. Also writtIng tests on Outthentic requires much less code. Also **much more detailed and accurate checks** of output are possible over those with bats. I suppose all of this can be of immense value for testing such a tools as parsers generating complex unstructured output. Imho this is where Outthentic seems more competitive.

scottchiefbaker:
I appreciate the effort, but I'm pretty happy with bats right now. I will look in to this more for future releases.

melezhik:
Hi @scottchiefbaker , sure. Let me know when you want to come back and consider this tool, I be glad to help you!

PR #282, opened 2018-01-25
Fix #277 Add Windows 10 Bash console installation note

### LINKED ISSUE #277: Problems installing on Windows 10
I don't understand installation procedure. I did the following:
system: **Windows 10 + built-in bash**

```
$ cd ~
$ curl https://raw.githubusercontent.com/so-fancy/diff-so-fancy/master/third_party/build_fatpack/diff-so-fancy -O
$ git config --global core.pager "diff-so-fancy | less --tabs=4 -RFX"
$ cd my-git-repo
$ git diff
diff-so-fancy | less --tabs=4 -RFX: diff-so-fancy: command not found
```

What should I do to get it working?

UnderHeavyDevelopment:


scottchiefbaker:
This isn't specific to Windows 10. Every platform is going to need the script in the path somewhere. Perhaps we expand the line: 

```
Installation is as simple as downloading the [diff-so-fancy](https://raw.githubusercontent.com/so-fancy/diff-so-fancy/master/third_party/build_fatpack/diff-so-fancy) script to a directory in your path. You can check which directories are in your path with `echo $PATH`.
```

Instead of separate instructions for Win10

UnderHeavyDevelopment:
Ok, looks nice. Changed pull request. This instruction update will be definitely helpful. If you prefer one commit, just let me know.

UnderHeavyDevelopment:
Following instruction, I changed pull request to `next` branch.

scottchiefbaker:
It's 12 commits right now... can you squash it down to one, since it's only a one line change.

scottchiefbaker:
Looks good. Thanks!

PR #285, opened 2018-02-11
Adding Hint for Windows users

Gegengift:


scottchiefbaker:
Instead of making this one big parenthesis addition, can we make the line an additional sentence. I think it would read cleaner. Something simple like

"Windows users may need to install the Window subsystem for Linux" and a link?

Gegengift:
Good idea

scottchiefbaker:
That's a large link... change the link to just the "Windows Subsystem for Linux" (remove the "install the") and I'll land this.

Gegengift:
done 👍

scottchiefbaker:
Next time we should look at squashing down the change to a single commit instead of three. That keeps the git history cleaner, and more readable. Since you're a first time contributor it's not a big deal.

Thanks for the help, we appreciate your hard work.

Gegengift:
How can I do this? Would I have to create a branch first and than squashed merge it back?

scottchiefbaker:
You don't need to make a branch. Here is a Stack Overflow [article](https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git) on it.

I already merged this, but in the future learning to squash commits will be a good skill to have. Especially on Github. Cheers!

Gegengift:
Thanks!

scottchiefbaker:
@Gegengift for what it's worth I would love a fresh set of eyes on our **README**. I've installed and used d-s-f 100 times so I'm not the best judge.

If you had time I'd love it if you could look at the whole README and tell me if there are other rooms for improvement. My goal is to make this as simple as possible to: install, setup, and use.

PR #289, opened 2018-03-01
pro-tips: Information about Zsh plugin/package providing diff-so-fancy

psprint:
As suggested I submit information about Zsh "package" (Zsh plugin, installed with some typical Zsh plugin manager).


scottchiefbaker:
Looks great! Thank you.

PR #294, opened 2018-03-27
Add config for file header ruler width (rulerWidth)

wesQ3:
I use `git log -p` frequently and I found the full-width file headers distracting when reading commits. I didn't want the separator between files within a commit to overshadow the separation of commits themselves. 

I think a shorter header width is better but naturally it should be an option!

Thanks for your work on this project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,13 @@ By default the separator for the file header uses Unicode line drawing character
 git config --bool --global diff-so-fancy.useUnicodeRuler false
 ```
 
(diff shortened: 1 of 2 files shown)


scottchiefbaker:
My first thought was that we don't want to shell out and get another config variable as it slows down startup. Looking at the code I forgot that we memoized the config so it won't be an extra hit. I'm fine with this code. Sorry for the delay.

PR #298, opened 2018-04-30
Make diff-so-fancy fancier. :)

dubstack:


scottchiefbaker:
This is so much fancier I can't say no.

PR #300, opened 2018-05-07
Match escape sequences wide enough to hold bold, truecolor fg, and truecolor bg.

### LINKED ISSUE #299: 24bit true color escapes in git `color.diff.*` config.
I haven't looked further yet, but on the surface it appears true color escapes are unrecognized. I use them in my git config to deviate a few very not-very-bold-but-hey-equals-bold-anyway colors in my theme, NeoSolarized, and I noticed `+` signs had returned to my diffs:

```diff
[color "diff"]
  new = "#859900" "#002b36" bold
+  #new = "#859900" "#002b36" bold
```

instead of:

```diff
[color "diff"]
  new = "#859900" "#002b36" bold
  #new = "#859900" "#002b36" bold
```

Reproduce by making the above git config change. I'll update if I dig deeper, otherwise dropping a note!

**EXTRA EXTRA**

This seems to work well:
```console
$ git config --global core.pager 'LESSOPEN="||-$(type -P diff-so-fancy)" && export LESSOPEN || unset LESSOPEN; exec less'
```

I can integrate/harden into docs if interested, but still need to confirm it works on massive input.

anthonyrisinger:
After a `git config --global color.diff.new '#859900 #002b36 bold'`, git will start
emitting escape sequences that look like `^[[1;38;2;133;153;0;48;2;0;43;54m`. Update
regex repetition to number of possible arguments in play (bold=1, fg=5, bg=5).

Fixes #299.

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -25,7 +25,7 @@ my $use_unicode_dash_for_ruler = git_config_boolean("diff-so-fancy.useUnicodeRul
 my $git_strip_prefix           = git_config_boolean("diff.noprefix","false");
 my $has_stdin                  = has_stdin();
 
-my $ansi_color_regex = qr/(\e\[([0-9]{1,3}(;[0-9]{1,3}){0,6})[mK])?/;
+my $ansi_color_regex = qr/(\e\[([0-9]{1,3}(;[0-9]{1,3}){0,10})[mK])?/;
 my $reset_color      = color("reset");
 my $bold             = color("bold");
 my $meta_color       = "";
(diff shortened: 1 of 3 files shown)


scottchiefbaker:
I'll gladly merge this if you can write a test for this as well. I can see something like this accidentally getting broken in the future and if we don't have a test we may not know.

scottchiefbaker:
Oh... one other thing, PRs should be against the `next` branch.

anthonyrisinger:
I can do that for sure, I'll update in a bit

anthonyrisinger:
@scottchiefbaker LMK if this works!

This bug is about _recognizing_ truecolor escapes, so it wasn't necessary to change any `git config color` entries in `test/test_helper/util.bash` because simply matching and removing the color is enough to verify proper function.

scottchiefbaker:
Tests look good, code looks good. I'm happy.

PR #304, opened 2018-07-18
Circle20

serkanh:


scottchiefbaker:
Any idea why this would make the Travis tests fail? Is that just coincidence?

nado:
I dont know if something changed at travis, but now you can see the results of shellcheck : https://travis-ci.org/so-fancy/diff-so-fancy/jobs/405166520#L1969
It complains about the use of `which` instead of `command -v`

kevgo:
You can now delete [circle.yml](https://github.com/so-fancy/diff-so-fancy/blob/master/circle.yml)

termux-io:
2f5a9475339e55281bd2ad07bf93312e881ecc8b``

PR #309, opened 2018-09-05
Unify build status badge look

kevgo:
This makes the CircleCI badge look similar to the other badges.


scottchiefbaker:
Looks good to me. More consistency is better.

PR #312, opened 2018-10-09
fix tests

stevemao:


.travis.yml
```diff
@@ -9,13 +9,6 @@ perl:
   - 5.18
   - 5.14
 
-matrix:
-  include:
-    - os: osx
-      osx_image: xcode7.3
```
stevemao: I'm not sure why we wanted to include an older version but removing this should fix the tests

PR #314, opened 2018-10-25
Fix so-fancy#311 Windows doesn't have '\dev\null'

webstech:
Update DiffHighlight to use  File::Spec->devnull() to avoid platform problems.

### DIFF
--- a/lib/DiffHighlight.pm
+++ b/lib/DiffHighlight.pm
@@ -4,6 +4,9 @@ use 5.008;
 use warnings FATAL => 'all';
 use strict;
 use Encode;
+use File::Spec;                                             # For devnull
+
+my $NULL = File::Spec->devnull();							# get correct value for unix/win
 
 # Highlight by reversing foreground and background. You could do
 # other things like bold or underline if you prefer.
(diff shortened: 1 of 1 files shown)


peff:
Seems reasonable. I think it is just `NUL` on Windows, but using `File::Spec` removes the need for us to do anything platform specific (and it should be available universally).

@webstech Do you have any interest in submitting this upstream to git.git? We do development on a mailing list, but you can use the PR->mail gateway at https://github.com/gitgitgadget/gitgitgadget/blob/master/README.md if that's easier.

scottchiefbaker:
`File::Spec` is a core module, so it should be installed everywhere. If @peff is OK with it I'm OK as well.

webstech:
@peff I will make the change in git.git.  Should this go against the master or maint branch?  I will change the comments to match the existing style.

peff:
@webstech `master` is fine, though it shouldn't matter in practice. That code hasn't changed between the two branches, so the patch will be the same either way (and since it's getting emailed eventually, the actual commit history isn't important; the maintainer will just apply it to the appropriate spot).

PR #315, opened 2018-10-31
Adds tip about scrolling issues in less on macOS

### LINKED ISSUE #246: Pager does not support scrolling via trackpad
I am using MacBook trackpad to scroll up and down the pages in the pager applications, such as `less` while viewing content that does not fit my terminal height.

Prior to configuring `diff-so-fancy`, I was able to run `git diff` and scroll through the diff using the trackpad. It does not work anymore after installing and configuring diff-so-fancy. It basically scrolls up to the output of prior programs by scrolling up the terminal. It should not be doing that.

Is it not working due to the diff-so-fancy's pager implementation, its piping to `less`, or because some configuration key is missing? 

Here is my config:

```
credential.helper=osxkeychain
core.pager=diff-so-fancy | less --tabs=4 -RFX
color.ui=true
color.diff-highlight.oldnormal=red bold
color.diff-highlight.oldhighlight=red bold 52
color.diff-highlight.newnormal=green bold
color.diff-highlight.newhighlight=green bold 22
color.diff.meta=227
color.diff.frag=magenta bold
color.diff.commit=227 bold
color.diff.old=red bold
color.diff.new=green bold
color.diff.whitespace=red reverse
core.repositoryformatversion=0
core.filemode=true
core.bare=false
core.logallrefupdates=true
core.ignorecase=true
core.precomposeunicode=true
```

corazzi:
Resolves #246

### DIFF
--- a/pro-tips.md
+++ b/pro-tips.md
@@ -55,3 +55,17 @@ zplug "zdharma/zsh-diff-so-fancy", as:command, use:bin/git-dsf
 zgen zdharma/zsh-diff-so-fancy
 ```
 
+#### Trackpad scrolling issues on macOS
+
+There are issues using the trackpad (or scrolling on a mouse) on macOS due to the version of `less` that is shipped with the operating system. To get around this, remove the `-X` flag on the pager configuration option and install `less` via Homebrew:
+
+```shell
+brew install less
+```
+
+You may need to refresh your `PATH` to use the new `less`:
+
+```shell
+hash -d less
+# or, hash -r
+```


pro-tips.md
```diff
@@ -55,3 +55,17 @@ zplug "zdharma/zsh-diff-so-fancy", as:command, use:bin/git-dsf
 zgen zdharma/zsh-diff-so-fancy
 ```
 
+#### Trackpad scrolling issues on macOS
+
+There are issues using the trackpad (or scrolling on a mouse) on macOS due to the version of `less` that is shipped with the operating system. To get around this, remove the `-X` flag on the pager configuration option and install `less` via Homebrew:
```
stevemao: I'd like to elaborate what "issues" they are. Maybe just link to #246?

pro-tips.md
```diff
@@ -55,3 +55,17 @@ zplug "zdharma/zsh-diff-so-fancy", as:command, use:bin/git-dsf
...
+
+```shell
+brew install less
```
OliverJAsh: Why is this needed? I don't see it mentioned in #246. Perhaps we should specify a reason?

pro-tips.md
OliverJAsh: Is it worth mentioning what we lose by removing `-X`? Reading #246 it sounds like a trade off. We have to decide between using `-X` for various other reasons, or not using it and having scroll work. We can't have both AFAICS, which is a shame.

scottchiefbaker:
I don't have OSX to test with. But all the bug reports we've received on this have been related to the built in less. Install GNU `less` solves the problem in every case. If you want "fancy" scrolling features beyond keyboard then the only way to get them is to install GNU `less`

OliverJAsh:
@scottchiefbaker I have mouse/trackpad scrolling working with macOS's built in `less`, by deactivating the `X` flag.

scottchiefbaker:
In all cases switching to GNU less has solved the issues that have cropped up. If people really want to stick with `less` that comes with OSX that's fine but there will be caveats because of known issues. This is about the fourth less/OSX issue that's cropped. For now my official stance is "if your on OSX and paging isn't working switch to GNU less" until we have a better solution that works cross platform.

OliverJAsh:
@scottchiefbaker In my testing just now, switching to GNU less (`brew install less`?) doesn't help. You must also deactivate the `X` flag.

MrChocolatine:
I was about to post an issue about this exact same problem, and I found all the similar issues.

I am on macOS Mojave and I also confirm `brew install less` does not solve the problem at all. However removing the `X` from the command `git config --global core.pager "diff-so-fancy | less --tabs=4 -RFX"`, provided in the [README](https://github.com/so-fancy/diff-so-fancy#usage), solves it.

OJFord:
MacOS Catalina 10.15.4 actually ships with GNU less, v487 (March 2017). (Brew provides v551, June 2019.)

I'm not sure of the history of versions it's had built-in, but I suspect it's more likely Apple bumped the version past one that fixed this bug, and that made it no longer necessary. Of course, if that's the case, people on older OSs would still need to install the newer version from homebrew, but that should be handled by the brew formula, since it's a package management issue really.

Just to double check - you definitely have Homebrew's 'cellar' ahead of `/usr/bin` on your `$PATH`, or else have supplied the absolute path in git config? Also, if Mojave -> Catalina really did switch from BSD to GNU less, brew may have install it as `gless` or similar. (There's typically a flag on install to tell it not to do that, or another dir called `gnubin` with unprefixed names that you can put on your PATH.)

MrChocolatine:
What do you want to double check ? I install `less` with `brew` so yes I get their last version available.

After a complete cleanup of `brew`, cleanup + remove the cache, I reinstalled `less` (version `551`) which gets correctly activated over the built-in `less`.
Then I reenabled the line `pager = diff-so-fancy | less --tabs=4 -RFX` in my gitconfig file.
Same problem again.

OJFord:
OK, yes, I can reproduce this, with both less v487 (shipped with macOS) and v551 (homebrew) via simply:
```
LESS= yes | less -X
```

And having read more about it, it's not even a less bug, it's `-X` doing exactly what it's supposed to do - disabling 'cup mode' (`man terminfo`) and filling the _current_ screen with the paged content, so of course scrolling back up goes 'past' the start, because you're still in you're shell, even if the current (still-running) command's output 'happens' to be filling the screen.

It's also **not macOS-specific**, or even really less-specific - if you disable switching to an alternate screen, your scroll buffer is going to include whatever came before still.

Personally I'd lean more towards 'pipe diff-so-fancy's output to whatever pager you want, and configure that however you want', but that's probably not very helpful to the 80% that want the damn thing to _just_work_.

In light of OliverJAsh's (18mo old, so I won't tag) https://github.com/so-fancy/diff-so-fancy/pull/315#discussion_r239416967 pointing out that `git` configures the `LESS` environment variable, if unset, to have the same `FRX` options we're suggesting, I wonder if a clean solution might just be to drop that, and leave it at `| less`, so it picks up `$LESS` if deliberately configured (or user can do that in git config), or else `-FRX` from git, and we don't have to take an opinion on correct scrolling behaviour or the relative importance of the resetting `-X` achieves. That 80% would still get `-FRX` (from git), and anyone who cares won't think it's more important to dsf which pager you use and how you can figure than it is, and can change it.

PR #320, opened 2019-01-16
Fix not working option --set-defaults

navrkald:
This pull req. fixes all options which contains dash inside string.


navrkald:
@scottchiefbaker I see now it has been already fixed in next branch but when will be next release? I need --set-defaults option for automatic installing of this awesome tool.
Thanks a lot.

scottchiefbaker:
I'm overdue to release the next branch. If I have time this afternoon I'll look at tackling it!

scottchiefbaker:
Version 1.2.5 is out now and contains this fix already.

PR #322, opened 2019-01-27
Fix #321: build fails on MacOS

### LINKED ISSUE #321: Needed: TravisCI / OSX Expert to figure out why tests are failing
Our tests are clean on every platform except OSX.

https://travis-ci.org/so-fancy/diff-so-fancy/jobs/481084220

I can't even figure out what it's complaining about. Can someone with more Travis experience than me take a look at why our tests are failing on OSX?

BernhardWebstudio:
A simple upgrade of the MacOS/XCode version is sufficient to get the builds working on MacOS too. I hope this is an acceptable solution. There is still an error reported by shellcheck ([SC2230](https://github.com/koalaman/shellcheck/wiki/SC2230)) but I guess it is not related to MacOS itself but rather the shellcheck version or configuration.

My second commit is not necessary to fix #321 , it just follows a best practice suggested by Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ perl:
 matrix:
   include:
     - os: osx
-      osx_image: xcode7.3
+      osx_image: xcode9.4
       language: generic
       perl: 5.24
   allow_failures:
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,11 +19,16 @@ matrix:
   allow_failures:
     - perl: "blead"
 
+addons:
+  homebrew:
+    packages:
+      - perl
+      - shellcheck
+      - cpanminus
+
 before_install:
   - git clone --depth 1 https://github.com/sstephenson/bats.git
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
-      brew update;
-      brew install shellcheck perl cpanminus;
       mkdir -p ~/perl5;
     fi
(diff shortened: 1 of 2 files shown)


scottchiefbaker:
THANK YOU SO MUCH FOR LOOKING IN TO THIS!

Any idea why `shellcheck` would error on Mac but not other platforms?

BernhardWebstudio:
No, I do not know why shellcheck sees the problem on MacOS only. I understand the problem, the documentation is clear, the lines are accurate, but I have no idea why the other Unix systems do not have the same problem. One could implement a fix by either follow the instructions in the shellcheck documentations or by disabling this specific error with the directive `# shellcheck disable=SC2230`. Or open an issue and ask the shellcheck authors why different nix Systems throw different errors.

scottchiefbaker:
`shellcheck` was a lot more relevant when d-s-f was mostly shell scripts (it's 90% Perl now). I'll probably just remove the `shellcheck` check all together. 

I just noticed this is a PR against `master` and it needs to be against `next` instead. If you can rework this to be against `next` I'll land the PR ASAP.

BernhardWebstudio:
I changed the base branch & fixed shellcheck, the tests succeed all now.

scottchiefbaker:
You sir... are a gentlemen and a scholar! Thank you so much for doing this.

This really needed fixing, and I don't have the expertise. Much appreciated.

PR #326, opened 2019-03-04
Fix for old Perl 5.8

jfcherng:
I just found that the current [d-s-f](https://raw.githubusercontent.com/so-fancy/diff-so-fancy/master/third_party/build_fatpack/diff-so-fancy) could not run on my old machine with Perl 5.8.8. And luckily  I found an old PR for the same problem. With this patch, it could be run.

References: https://github.com/so-fancy/diff-so-fancy/pull/177

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -768,7 +768,10 @@ sub color {
 
 	# Some predefined colors
 	my %color_map = qw(red 160 blue 21 green 34 yellow 226 orange 214 purple 93 white 15 black 0);
-	$str =~ s|([A-Za-z]+)|$color_map{$1} // $1|eg;
+	# For old Perl which cannot use "//", we sometimes use "||" to replace it.
+	# But unfortunately the "black" here is mapped to 0.
+	# Thus, we can only expand the "//" syntax sugar to what it means exactly.
+	$str =~ s/([A-Za-z]+)/(defined $color_map{$1}) ? $color_map{$1} : $1/eg;
 
 	# Get foreground/background and any commands
 	my ($fc,$cmd) = $str =~ /(\d+)?_?(\w+)?/g;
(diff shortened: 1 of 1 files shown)


scottchiefbaker:
Supporting PHP 5.8 isn't a huge priority for d-s-f due to it's age. If we could support 5.8 and newer version I'd definitely support that.

That said, the change you made to line 771 would break backwards compatibility for "black" as it maps to 0, which would then not evaluate correctly.

diff-so-fancy
```diff
@@ -768,15 +768,15 @@ sub color {
 
 	# Some predefined colors
 	my %color_map = qw(red 160 blue 21 green 34 yellow 226 orange 214 purple 93 white 15 black 0);
-	$str =~ s|([A-Za-z]+)|$color_map{$1} // $1|eg;
+	$str =~ s/([A-Za-z]+)/$color_map{$1} || $1/eg;
```
scottchiefbaker: If someone requests "red_on_black" or "black_on_yellow" this would break.

"black" evaluates to `0` which then would not pass the `||` check. This would result in "black" making it through this check unaltered and cause other side effects.

diff-so-fancy
```diff
@@ -768,15 +768,15 @@ sub color {
...
 
 	# Get foreground/background and any commands
 	my ($fc,$cmd) = $str =~ /(\d+)?_?(\w+)?/g;
 	my ($bc)      = $str =~ /on_?(\d+)/g;
 
 	# Some predefined commands
 	my %cmd_map = qw(bold 1 italic 3 underline 4 blink 5 inverse 7);
-	my $cmd_num = $cmd_map{$cmd // 0};
+	my $cmd_num = $cmd_map{$cmd || 0};
```
scottchiefbaker: This one is OK because none of the commands resolve to `0` so it won't affect anything.

diff-so-fancy
jfcherng: > If someone requests "red_on_black" or "black_on_yellow" this would break.
> 
> "black" evaluates to `0` which then would not pass the `||` check. This would result in "black" making it through this check unaltered and cause other side effects.

Ah! Nice catch. But I am not sure that we are willing to expand the `//` syntax sugar or not...
```perl
$str =~ s/([A-Za-z]+)/(defined $color_map{$1}) ? $color_map{$1} : $1/eg;
```
with some comments about why we wrote it like this.

PR #330, opened 2019-04-26
feat: readme install instructions conditional core.pager

mightyiam:


mightyiam:
When I wrote that it's tested, I meant in Linux. It would be nice to have a confirmation regarding Windows and Mac.

scottchiefbaker:
What does this fix?

mightyiam:
@scottchiefbaker I don't always have `diff-so-fancy` in my `$PATH`. Only when I use my choice of a Node.js version manager to select a Node.js version, only then do I have `diff-so-fancy` in my `$PATH`. When I start my shell, no Node.js version is automatically selected. I have `diff-so-fancy` installed via npm as a global package. Let me show you:

```
➜  ~ git:(master) ✗ which diff-so-fancy
diff-so-fancy not found
➜  ~ git:(master) ✗ nvs 
PATH += ~/.nvs/node/8.15.1/x64/bin
➜  ~ git:(master) ✗ which diff-so-fancy
/home/shahar/.nvs/node/8.15.1/x64/bin/diff-so-fancy
➜  ~ git:(master) ✗ 
```

This change supports that use case.

scottchiefbaker:
This really complicates the alias, and seems to only fix a corner case. I'm glad you found a fix, but I think it's appropriate for the larger d-s-f audience.

Issue #337, opened 2019-06-24
Travis tests are failing on OS X


scottchiefbaker:
Builds have been failing on OS X for a while now and I have no idea why.

@BernhardWebstudio any way I can get you to take another look at our Travis config again?

BernhardWebstudio:
Well, on _master_ they are because there it is still the old `.travis.yml`. On _next_, the main issue seems to be the error message `Your Homebrew is outdated. Please run _brew update_.`, which prevents perl, cpanminus etc. from being installed. I will submit a PR shortly...

scottchiefbaker:
Fixed after merge!

PR #338, opened 2019-06-24
Fix MacOS Travis Tests

BernhardWebstudio:
Fix MacOS Travis Tests by requiring Homebrew to manually update before installing its dependencies. 

This fix still leaves me wondering how this option is one at all, when there is no way to install dependencies with an outdated brew database, meaning without this option.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,14 +25,15 @@ addons:
       - perl
       - shellcheck
       - cpanminus
+    update: true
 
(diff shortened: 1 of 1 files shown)


BernhardWebstudio:
Ah, and fixes #337

scottchiefbaker:
That was an easy fix!

Thanks so much!

PR #339, opened 2019-06-30
support coloring hunk header

cstrahan:
This brings diff-so-fancy inline with the screenshot: before this change, the hunk headers were unstyled (maybe this behavior was lost in a past revision? I'd do a bisect, but the old versions of d-s-f don't work with my version of `git`).

Also, rather than always making the last function bold, leave that to the user to configure (though make the default fallback be bold when the user has not configured `color.diff.func`).

I have no strong feelings about the default color choice for `color.diff.frag`, so feel free to change that.

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -249,7 +249,8 @@ sub do_dsf_stuff {
 
 			# Last function has it's own color
 			my $last_function_color = get_config_color("last_function");
-			print "@ $last_file_seen:$start_line \@${bold}${last_function_color}${remain}${reset_color}\n";
+                        my $frag_color = get_config_color("frag");
+                        print "${frag_color}@ $last_file_seen:$start_line \@${reset_color}${last_function_color}${remain}${reset_color}\n";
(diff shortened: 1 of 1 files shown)


cstrahan:
Oh, I just realized that I wasn't forcing git to use color when piping to `diff-so-fancy`. Had I done this, I would find that the colors were preserved:

```bash
git -c color.ui=always diff | ~/bin/diff-so-fancy | less --tabs=4 -RFXm
```

Perhaps it would still be worthwhile to remove the `${bold}` from the last function, though, so users can have control over that.

cstrahan:
For posterity, I wanted a git command by which I could explicitly opt into having `git diff` filtered through d-s-f. This is what I've been using (called `fdiff` as in "fancy diff"):

```gitconfig
# in my ~/.gitconfig:
[alias]
  fdiff = -c color.ui=always -c core.pager=\"diff-so-fancy | less --tabs=4 -RFXm\" diff
```

PR #340, opened 2019-06-30
Don't force bold for last function

cstrahan:
Also, reset color so e.g. bold from the hunk header doesn't carry over to the last function.

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -249,7 +249,7 @@ sub do_dsf_stuff {
 
 			# Last function has it's own color
 			my $last_function_color = get_config_color("last_function");
-			print "@ $last_file_seen:$start_line \@${bold}${last_function_color}${remain}${reset_color}\n";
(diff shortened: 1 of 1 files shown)


scottchiefbaker:
I'm fine with this in theory... can you elaborate on exactly what this fixes before I land it?

Just want to be 100% clear I understand what's going on here.

cstrahan:
There are two problems with the current setup that this PR addresses:

1. Some users may not want the "last function" to be `bold`, and may be surprised to find that, despite setting just `magenta`, they effectively get `bold magenta`.
2.  If the user has set `color.diff.frag` (the style for hunk headers -- referred to as "last file" throughout the `diff-so-fancy` source) to include `bold` or `reverse`, those effects will currently get inherited during the display of the last function.

This fixes both issues by replacing the `${bold}` with `${reset_color}`, and instead making the default style (that is, if the user hasn't explicitly set `color.diff.func` in their gitconfig) be bold (this is in accordance with the pattern set for how the other styles are handled elsewhere in `diff-so-fancy`).

PR #342, opened 2019-08-05
avoid line wrapping in less

ryenus:
When line numbers are shown in less, the horizontal ruler would wrap
around if line wrapping is turned on, there're 2 options:
1. turn off line wrapping in less, using `-S`
2. decrease horizontal_rule width, e.g.: `$width -= 8`,
   to accommodate line numbers up to 4 digits.

I added `-S`, but It should be fine to use either here.


scottchiefbaker:
We're not showing line numbers with less, I'm confused how this is coming up.

ryenus:
Well, line numbers can be turned on by either defining the `LESS` environment variable (e.g., `LESS=-N`), or by typing `-N` directly when inside less.

Then, with line numbers turned on, plus line wrapping is ON by default, the horizontal ruler would wrap around since it has full width as that of the terminal window.

PR #356, opened 2019-11-10
docs: add 'color.diff.func' color to README

phil-blain:
Two small improvements to the documentation:

- `git --no-pager` can be abbreviated `git -P` in Git 2.18 or later,
so mention this in pro-tips.md

- The color for hunk headers (`color.diff.func`) is hardcoded in the
script, and is not mentioned in the "Improved colors for the
highlighted bits" section of the README. Add it there so that function
headers with or without diff-so-fancy use the same color configuration.

I based this branch on master since it appears next is behind master. I will rebase on next if required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ git config --global color.diff-highlight.newHighlight "green bold 22"
 
 git config --global color.diff.meta       "11"
 git config --global color.diff.frag       "magenta bold"
+git config --global color.diff.func       "146 bold"
 git config --global color.diff.commit     "yellow bold"
 git config --global color.diff.old        "red bold"
 git config --global color.diff.new        "green bold"


phil-blain:
@scottchiefbaker did you have a chance to look at these changes ? 

By the way travis is failing for an unrelated reason:
https://travis-ci.org/so-fancy/diff-so-fancy/jobs/609991970#L105

~~~bash
$ cpanm --quiet --notest --local-lib=~/perl5 local::lib
/Users/travis/.travis/functions: line 109: cpanm: command not found
The command "cpanm --quiet --notest --local-lib=~/perl5 local::lib" failed and exited with 127 during .~~~

scottchiefbaker:
I'm not a huge fan of documenting the `-P` thing. Are we going to document the alternate form of every param? I'm a **big** fan of the long version anyway because it's much more descriptive and doesn't require me to load the man page in six months to remember what it does.

The other change is more interesting, and I would consider merging that. I will look in to it further when I have more time.

And yes, pull-requests should be based on **next** not master.

phil-blain:
Hi @scottchiefbaker !

I removed the `-P` bit and kept just the `color.diff.func` bit. Also I rebased on `next` and changed the target branch for the PR.

Cheers!

PR #358, opened 2019-11-23
Add info about using dsf as a "normal" diff tool

### LINKED ISSUE #328: Docs: Emphasize that diff-so-fancy is usable as a general diff tool
I still use Subversion a lot and diff-so-fancy is great as a tool for Subversion diffs, too.  Heck, any diffs.  Currently, it sounds like it's a git-only tool.

martijnengler:
diff-so-fancy is great for git of course, but it's also very nice to use
for just normal diffs between two files

Fixes #328

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ diff -u file_a file_b | diff-so-fancy
 You could, of course, create a function for it too. In your `~/.bashrc` or `~/.zshrcz` for example:
 
 ```shell
-function dsf() { diff -u $1 $2 | diff-so-fancy; }
+function dsf() { diff -u "$1" "$2" | diff-so-fancy; }
 ```
 
 And than you can just type `dsf file_a file_b` to get nice looking diff output.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,6 +18,26 @@ Windows users may need to install [MinGW](https://sourceforge.net/projects/mingw
...
+Use `-u` with diff to for unified output, and pipe the output to `diff-so-fancy`:
+
+```shell
+diff -u file_a file_b | diff-so-fancy
+```
+
+You could, of course, create a function for it too. In your `~/.bashrc` or `~/.zshrcz` for example:
+
+```shell
+function dsf() { diff -u $1 $2 | diff-so-fancy; }
```
wren: This should have quotes around the `$1` and `$2` in order to handle filenames with spaces.

scottchiefbaker:
I'll merge this and massage the wording/order a little bit. Thanks.

PR #379, opened 2020-08-01
Add pro-tip about hg diff

srsudar:
As requested on #186.

LMK if you want some style changes, or feel free to edit before merging.

Issue #186, opened 2016-07-20
Mercurial?


Daniel15:
Do you have instructions on how to configure this project to be used with Mercurial rather than Git?

OJFord:
That would be dependent upon mercurial diffs being line-compatible with git's; I don't know if that's the case?

I've never used mercurial (or, therefore, tested this) - but you could try:
```ini
[extensions]
hgext.extdiff =
[extdiff]
cmd.dsf = diff-so-fancy | less --tabs=4 -RFX
```

Daniel15:
Thanks for the info! I just tried that, which gave me this:
```
sh: diff-so-fancy | less --tabs=4 -RFX: command not found
```

Seems like it's trying to run the whole thing as a command, rather than treating `|` as a pipe.  I tried removing the pipe to `less` and just doing:
```
cmd.dsf = diff-so-fancy
```

But then it just sat there forever, without doing anything.

I ended up just creating a `hg-diff-fancy` shell alias, which is working fine for me:
```
hg-diff-fancy='hg diff --color=always | diff-so-fancy | less -FR'
```

OJFord:
> Seems like it's trying to run the whole thing as a command

Hm, actually the example at [the bottom of this page](https://www.mercurial-scm.org/wiki/ExtdiffExtension) does make it seem like pipes aren't supported for some reason.

> I tried removing the pipe ... But then it just sat there forever

That would suggest that git/hg diffs aren't compatible; so it chokes on the input. So it's interesting that you were able to create an
> hg-diff-fancy shell alias, which is working fine
I'm not sure what's going on there then.

Still, at least it works I suppose! I might look into this at some point, see if I can make it work without needing a shell-level alias so we can document it as known-working (if unsupported and may break) with mercurial.

Daniel15:
I think Mercurial diffs are compatible with Git diffs (that is, they're in the same format). Not sure if that's the default configuration, but it's how it seems to be configured internally at Facebook.

OJFord:
Looked into this a little bit, it seems both are based on GNU's "Unified Format", which makes sense. Git seems to extend it a little bit though, so if you run into issues with your current command, you may want to [try adding `--git` to `hg diff`](https://www.mercurial-scm.org/wiki/GitExtendedDiffFormat).

srsudar:
[...]

PR #382, opened 2020-08-27
Add pro-tip about hg diff

srsudar:
As requested on #186, then botched on #379.


scottchiefbaker:
Looks good! Thanks for the addition.

PR #387, opened 2020-11-06
Don't include ansi color codes in file1 and file2

ericbn:
when processing renamed files. Use the same `(.+?)(\e|\t|$)` regexp used in other similar parts of the code. The motivation behind this change is to be able to process the file1 and file2 strings (e.g. get their length) without the interference of the extra ansi color codes. Also, the resulting d-s-f output is smaller with this change (34 bytes less for each "renamed: X to Y" line).

Screen Shot 2020-11-03 at 08 36 35

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -318,10 +318,10 @@ sub do_dsf_stuff {
 			}
 
 			my $next    = shift(@$input);
-			my ($file1) = $next =~ /rename from (.+)/;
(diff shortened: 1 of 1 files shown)


scottchiefbaker:
Interesting... 

Looking at this code it looks clean. Running the tests come back fine. I'm 95% sure this is going to get merged. I'm going to run with it on my local code base for a week or two to see if I notice anything weird. Assuming I don't I'll merge this after that.

Thanks for the PR.

PR #390, opened 2020-12-07
Add patch mode for use in `git add --patch`

### LINKED ISSUE #35: Git add --patch does not use diff-so-fancy

wren:
Fixes #35

This PR does two things:
1. Fix a bug where a color reset code was sometimes adding an extra line to diff-so-fancy output. In regular diffs, this bug had no impact, but was causing a lot of headahces when trying to use diff-so-fancy with `git add --patch`.
2. Add a `--patch` flag to make diff-so-fancy compatible with `git add --patch`. When this flag is active, diff-so-fancy will pad its output with empty lines in order to meet a git requirement that the diff be the same number of lines in length.

Due to the nature of the changes (only adding empty lines), I'm not sure how to test this using bats (it can't detect blank lines). So, any advice on this would be very welcome.

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -714,9 +714,13 @@ diff -u one.txt two.txt | diff-so-fancy  # Use d-s-f on unified diff output
 
 diff-so-fancy --colors                   # View the commands to set the recommended colors
 diff-so-fancy --set-defaults             # Configure git-diff to use diff-so-fancy and suggested colors
+diff-so-fancy --patch                    # Use diff-so-fancy in patch mode (interoperable with `git add --patch`)
 
 # Configure git to use d-s-f for *all* diff operations
-git config --global core.pager \"diff-so-fancy | less --tabs=4 -RFX\"\n";
+git config --global core.pager \"diff-so-fancy | less --tabs=4 -RFX\"
+
+# Configure git to use d-s-f for `git add --patch`
+git config --global interactive.diffFilter \"diff-so-fancy --patch\"\n";
 
 	return $out;
 }
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -543,13 +543,20 @@ sub strip_leading_indicators {
 	if ($manually_color_lines) {
 		if (defined($5) && $5 eq "+") {
 			my $add_line_color = get_config_color("add_line");
-			$line              = $add_line_color . $line . $reset_color;
+			$line              = $add_line_color . insert_reset_at_line_end($line);
 		} elsif (defined($5) && $5 eq "-") {
 			my $remove_line_color = get_config_color("remove_line");
-			$line                 = $remove_line_color . $line . $reset_color;
+			$line                 = $remove_line_color . insert_reset_at_line_end($line);
 		}
 	}
 
+	return $line;
+}
+
+# Insert the color reset code at end of line, but before any newlines
+sub insert_reset_at_line_end {
+	my $line = shift();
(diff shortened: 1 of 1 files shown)


wren:
@scottchiefbaker I had started down the road of keeping track of how many lines needed to be added, and what mode we are in, and all that, and it was getting complicated. Then I realized that we don't really need any of that, and we just need to print an extra line whenever we consume a line without any output. Turns out there are only three spots where that happens.

Not sure if this in okay approach. But I'm happy to change anything, so just let me know!

scottchiefbaker:
This looks good. How do you configure git to use this mode for `--patch` stuff?

diff-so-fancy
```diff
@@ -178,6 +186,10 @@ sub do_dsf_stuff {
 
 			$last_file_seen =~ s|^\w/||; # Remove a/ (and handle diff.mnemonicPrefix).
 			$in_hunk = 0;
+			if ($patch_mode) {
+				# we are consuming one line, and the debt must be paid
```
scottchiefbaker, line 190: Hah! I like this

wren:
To configure git to use this, run this command: 

```sh
git config --global interactive.diffFilter 'diff-so-fancy --patch'
```

Or put this into your git config
```gitconfig
[interactive]
  diffFilter = diff-so-fancy --patch
```

That is in addition to configuring diff-so-fancy for use with diffs. So, a config that uses both looks like this:
```gitconfig
[core]
  pager = diff-so-fancy | less --tabs=4 -RFX
[interactive]
  diffFilter = diff-so-fancy --patch
```

scottchiefbaker:
Wow!

1) Thank you so much for your time and energy tracking this down. It's been a long standing issue (February of 2016!) that has affected a lot of users.
2) Your code looks great... super simple, easy to read, and using the same style as the rest of the code. Bonus points for your attention to detail on this
3) I did some testing on my local code base and it `Just Works (tm)` out of the box!

I'm tempted to merge this to `next` and let people start testing. Assuming things go smoothly for everyone I'll go ahead and push out a new release after we get enough thumbs up from testers.

The only thing I would need know would be a documentation update to explain the `interactive` filter so people know how to use it. I'm thinking just a second example line in the **usage** section would be sufficient.

diff-so-fancy
```diff
@@ -51,6 +52,13 @@ if ($args->{color_on}) {
 	$color_forced         = 1;
 }
 
+# `git add --patch` requries our output to match the number of lines from the
```
webstech, line 55: `requries` -> `requires`. thanks.

wren:
@scottchiefbaker Happy to help!

Also, I updated the usage section to reflect the new flag. Just let me know if you need anything else.

diff-so-fancy
wren: Fixed in [ed24bc](https://github.com/so-fancy/diff-so-fancy/pull/390/commits/ed24bc15b1e789f7c7304898016f255588e95c96).

scottchiefbaker:
This looks good enough for now. I'm going to merge it in to `next` and let some people test it out. If things test well then I'll do a formal release.

Thanks for your work on this.

PR #391, opened 2020-12-07
(#35) Add --pager option for diff.interactiveFilter

### LINKED ISSUE #35: Git add --patch does not use diff-so-fancy

ulwlu:
Fix #35 

## This PR changes below.
- If user supplies `--pager` option, dfs won't custom meta lines.

ss 3

## Usecase
```
[interactive]
	diffFilter = diff-so-fancy --pager
```

## Motivation
I'm a contributor of [git-delta](https://github.com/dandavison/delta) and solved the same problem. I found #35 issue today and did a same approach.

## Remarks

I'd like to add a test. In order to test, we need to confirm each outputs below have same line-number.

```
# for input
git diff-files -p --no-color

# for output
git diff-files -p --color | diff-so-fancy --pager
```

However, I'm not familiar with bats, and bats seems not to able to achieve .... Is there any good suggestion to add a test?

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -41,6 +41,7 @@ my $in_hunk           = 0;
 my $columns_to_remove = 0;
 my $is_mercurial      = 0;
 my $color_forced      = 0; # Has the color been forced on/off
+my $pager             = 0; # Is diff.interactiveFilter mode
 
 # We try and be smart about whether we need to do line coloring, but
 # this is an option to force it on/off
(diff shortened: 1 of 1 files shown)


ulwlu:
Sorry, somebody seems creates same PR 14 hours ago.
I'm gonna close this.

Issue #278, opened 2018-01-03
Remove dependence on `git config --get-color` (implement Git coloring internally)


scottchiefbaker:
Right now d-s-f shells out to Git 5+ times to get color configuration information. This is needed to translate Git color strings into ANSI escape sequence. This process slows down startup time.

Ideally we can read the entire Git config once with `git config --list` and process the entries one by one. We'll need a list of the Git supported strings and what ANSI codes they should map too. 

Example strings: "red bold", "green bold 22", "red reverse", "yellow"

scottchiefbaker:
@peff could you provide a starting point on the Git side about where this information is stored? Or a list of what strings Git accepts?

peff:
The interesting entry point is [https://github.com/gitster/git/blob/4a4ac8367882698ac607169f65964c344f74d14c/color.c#L202](`color_parse_mem`), which in turn relies on `parse_color` and `parse_attr`. t's _pretty_ straightforward for porting, I think.

I do still think you'd be better served by a `git config --stdin` feature. I had somebody working on it during a hackfest in November, but unfortunately they seem to have stalled halfway through. :cry:

scottchiefbaker:
Can you elaborate on `git config --stdin`?

peff:
I mean that ideally you could do something like:

```
git config --stdin <<\EOF
key=color.diff-highlight.oldnormal
type=color

key=color.diffhighlight.oldhighlight
type=color
...etc...
EOF
```

and get output something like:

```
key=color.diffhighlight.oldnormal
value=[some ansi codes]
...etc...
```

all with a single invocation of `git config`. But that feature doesn't exist yet.

scottchiefbaker:
At startup diff-so-fancy shells out to `git --config` [four times](https://github.com/so-fancy/diff-so-fancy/blob/master/lib/DiffHighlight.pm#L15) to read color data from the config file. This is slowest part of startup and can make d-s-f feel sluggish. It has been a personal pet peeve of mine for a while, and I would like to speed up the startup time.

I ran d-s-f five times with the parsing code in place and got an average of 81 ms to show a simple diff. Then I hardcoded `color_config()` to always return 4 and re-ran the tests and I got an average of 46 ms to show the same diff.  In other words, shelling out to check colors almost doubles startup time.

All tests were run with a simple:

    time git diff

scottchiefbaker:
@peff did `git config --stdin` ever get implemented?

peff:
[...]

PR #398, opened 2021-02-20
Draw ruler and box around headers

ericbn:
Rulers are drawn around the first level of headers, and boxes around the second level of headers, if any.

The `diff-so-fancy.rulerWidth` config still applies, but only to the rulers. Boxes have just the size of the text inside them. E.g. `git show` and `git log` have two levels of headers: (1) commit, and (2) meta (e.g. `modified: diff-so-fancy`). `git diff` only has one level of headers: meta.

This allows for a better hierarchical view of commits specifically, with commit lines having a configurable wider ruler around them and meta lines ("children" of commits) having a shorter box around them.

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -27,11 +27,37 @@ my $ruler_width                = git_config("diff-so-fancy.rulerWidth", undef);
 my $git_strip_prefix           = git_config_boolean("diff.noprefix","false");
 my $has_stdin                  = has_stdin();
 
+my $box_horizontal;
+my $box_vertical;
+my $box_down;
+my $box_up;
+# BOX DRAWINGS LIGHT HORIZONTAL http://www.fileformat.info/info/unicode/char/2500/index.htm
+# BOX DRAWINGS LIGHT VERTICAL https://www.fileformat.info/info/unicode/char/2502/index.htm
+# BOX DRAWINGS LIGHT DOWN AND LEFT https://www.fileformat.info/info/unicode/char/2510/index.htm
+# BOX DRAWINGS LIGHT UP AND LEFT https://www.fileformat.info/info/unicode/char/2518/index.htm
+if ($use_unicode_dash_for_ruler && should_print_unicode()) {
+	# $box_horizontal = Encode::encode('UTF-8', "\x{2500}");
+	$box_horizontal = "\xE2\x94\x80";
+	# $box_vertical = Encode::encode('UTF-8', "\x{2502}");
+	$box_vertical = "\xE2\x94\x82";
+	# $box_down = Encode::encode('UTF-8', "\x{2510}");
+	$box_down = "\xE2\x94\x90";
+	# $box_up = Encode::encode('UTF-8', "\x{2518}");
+	$box_up = "\xE2\x94\x98";
+} else {
+	$box_horizontal = "-";
+	$box_vertical = "|";
+	$box_down = ".";
+	$box_up = "'";
+}
+
 my $ansi_regex = qr/\e\[([0-9]{1,3}(;[0-9]{1,3}){0,10})[mK]/;
 my $ansi_color_regex = qr/(${ansi_regex})?/;
 my $reset_color      = color("reset");
 my $bold             = color("bold");
 my $meta_color       = "";
+my $commit_color     = "";
+my $min_header_level = 2;
 
 # Set the diff highlight colors from the config
 init_diff_highlight_colors();
@@ -159,13 +185,17 @@ sub do_dsf_stuff {
 
 		#######################################################################
 
-		####################################################################
-		# Look for git index and replace it horizontal line (header later) #
-		####################################################################
-		if ($line =~ /^${ansi_color_regex}index /) {
-			# Print the line color and then the actual line
+		########################
+		# Look for commit line #
+		########################
+		if ($line =~ /^${ansi_color_regex}(commit [0-9a-f]{40}.*)$/) {
+			$commit_color = $1 || get_config_color("commit");
+			print_header(1,$commit_color,$4);
+		######################
+		# Look for git index #
+		######################
+		} elsif ($line =~ /^${ansi_color_regex}index /) {
 			$meta_color = $1 || get_config_color("meta");
-
 			# Get the next line without incrementing counter while loop
 			my $next = $input->[0] || "";
 			my ($file_1,$file_2);
@@ -185,9 +215,7 @@ sub do_dsf_stuff {
 			}
 
 			if ($file_1 && $file_2) {
-				print horizontal_rule($meta_color);
-				print $meta_color . file_change_string($file_1,$file_2) . "\n";
-				print horizontal_rule($meta_color);
+				print_header(2,$meta_color,file_change_string($file_1,$file_2));
 			}
 		#########################
 		# Look for the filename #
@@ -198,7 +226,6 @@ sub do_dsf_stuff {
 			# Mercurial looks like: diff -r 82e55d328c8c hello.c
 			if ($4 eq "-r") {
 				$is_mercurial = 1;
-				$meta_color = get_config_color("meta");
 			# Git looks like: diff --git a/diff-so-fancy b/diff-so-fancy
 			} else {
 				$last_file_seen = $5;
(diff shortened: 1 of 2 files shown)


ericbn:
This is how it looks like:

## Before / After

Screen Shot 2021-02-19 at 20 23 56

test/test_helper/util.bash
```diff
@@ -13,9 +13,9 @@ set_env() {
 
 # applying colors so ANSI color values will match
 # FIXME: not everyone will have these set, so we need to test for that.
-git config color.diff.meta "227"
+git config color.diff.meta "yellow bold"
```
ericbn, line 16: Tests expect the "meta" color to be `\e[1;33` (yellow bold).

diff-so-fancy
```diff
@@ -515,7 +509,7 @@ sub get_less_charset {
 		}
 	}
 
-	return ();
+	return ("", "");
```
ericbn, line 512: diff-so-fancy was failing with `Use of uninitialized value $less_charset in pattern match (m//) at diff-so-fancy line 524` in the case where `()` was being returned to should_print_unicode.

diff-so-fancy
```diff
@@ -226,20 +242,7 @@ sub do_dsf_stuff {
...
 			}
 
-			# Print out the top horizontal line of the header
-			print $reset_color;
-			print horizontal_rule($meta_color);
-
-			# Mercurial coloring is slightly different so we need to hard reset colors
-			if ($is_mercurial) {
-				print $reset_color;
-			}
```
ericbn, line 236: horizontal_rule was already printing $reset_color at the end of the line, so an extra $reset_color here was redundant.

diff-so-fancy
```diff
@@ -145,13 +164,15 @@ sub do_dsf_stuff {
...
+		# Look for commit line #
+		########################
+		if ($line =~ /^${ansi_color_regex}commit [0-9a-f]{40}/) {
+			print_commit_box($line);
 		####################################################################
 		# Look for git index and replace it horizontal line (header later) #
 		####################################################################
-		if ($line =~ /^${ansi_color_regex}index /) {
-			# Print the line color and then the actual line
-			$meta_color = $1 || get_config_color("meta");
```
ericbn, line 153: This was causing the value of $meta_color to actually be the value of $1 (the capture group of the regexp) instead of the value from the configuration.

scottchiefbaker: This is by design... to get a config color requires shelling out to bash (slow). So we attempt to get the color from the diff output first, and if it's not there, falling back to checking the config.

scottchiefbaker:
Several of the things you fixed are already fixed on `next`. All PR should be based against that branch. Can you redo this PR to target the [correct branch](https://github.com/so-fancy/diff-so-fancy#contributing).

ericbn:
Hi @scottchiefbaker, sure! Let me rebase it.

ericbn:
Done. The changes depend on eb4a5e7 (Don't include ansi color codes in file1 and file2, from #387), which was merged to master, but not to next (ooops, sorry, I didn't follow the [contributing](https://github.com/so-fancy/diff-so-fancy#contributing) steps again there!). So I cherry cherry-picked the commit here.

diff-so-fancy
ericbn: Got it! We lost this optimization as I refactored this code into the print_file_change_box function. I can bring the optimization back. Please let me know if you rather have it back.

ericbn:
@scottchiefbaker, I apologize about the confusion around the fix for the "Use of uninitialized value $less_charset in pattern match" error. I didn't notice it was already fixed in the `next` branch. I've just removed the fix that I suggesting for this same error in this PR.

The only failure still left in this PR is in the macOS build, which seems to have been fixed in #403, but was merged to `master` instead of to `next`.

scottchiefbaker:
I've been hemming and hawing over this issue for a while. Personally I don't think I would use, but I can definitely some people in the community wanting it. The issue then becomes that we need to make this a configuration option to allow people to opt in/out of this feature. 

Right now every time d-s-f starts up we shell out to `git config --list` for every config item. This works, but it's slow, and it only get slower as we add more options. I need to ponder how best we can add this feature without slowing down the startup of d-s-f. It's already slower than I would like it.

ericbn:
Cool. I see d-s-f is already caching the git config output in `$static_config`. But it parses it once for each call of `sub git_config`. So maybe the parsed result can be cached instead, to avoid that extra redundant parsing.

Besides that, the code also uses `DiffHighlight::color_config`, and that can be replaced by respective calls to `sub git_config`, which takes advantage of the cached config. `DiffHighlight::color_config` on the other hand, [will call git config each time it's called](https://github.com/git/git/blob/bf949ade81106fbda068c1fdb2c6fd1cb1babe7e/contrib/diff-highlight/DiffHighlight.pm#L140-L144), as you mentioned.

Also, I'm doing the following in the only config-related code that I've added:
```
		if ($line =~ /^${ansi_color_regex}commit [0-9a-f]{40}/) {
			$commit_color = $1 || get_config_color("commit");
			print_commit_box($line);
```
meaning it's first tries to use the color matched in the regexp, using the same optimization that you already pointed out before.

So there's room to make it even more optimized than what it was before this PR.  :- )

Are you benchmarking d-s-f with any specific script?

scottchiefbaker:
@ericbn I just landed some commits to handle caching of `git_config()` that  might speed things up a bit. They're on `next` take a look and we can build from there.

scottchiefbaker:
@ericbn I updated #278 with some more details. If we can get that figured out, landing this will be a no-brainer.

scottchiefbaker:
@ericbn I'm circling back to this now that I got our startup issues figured out. I removed the `Encode` module all together because it was slowing down startup. I see you reference it here... I think all of those encode calls could be hardcoded. That's what I did with the horizontal rule character.

ericbn:
Hi @scottchiefbaker. I'm loving all the recent optimizations. I've just rebased the PR commit and updated it to hardcode the UTF-8 encodings. Thanks!

scottchiefbaker:
Does this kick in if there is a commit line, or all the time? Seems like this is a `git show` thing, not a `git diff`?

ericbn:
@scottchiefbaker, correct: the "commit lines" show up in `git show` and `git log`, but not in `git diff`. The `git diff` only has the "meta lines", for which I'm changing the formatting to a tighter box that goes exactly around the text width. And all the formatting is the same regardless of the output that is being shown.

Do you want me to change it so the "top hierarchical lines" (commits in show and log, meta in diff) will always show with the wide lines, and the "bottom hierarchical lines" (commits in show and log, N/A in diff) will always show with the tighter box?

ericbn:
@scottchiefbaker, I've updated the PR so the behavior with `git diff` is unchanged (only ruler is drawn around each meta line). For `git show` and `git log` it will print the ruler around commit lines, and a box around meta lines, to distinguish between the different header levels (like in the [screenshot posted above](https://github.com/so-fancy/diff-so-fancy/pull/398#issuecomment-782520586)).

ericbn:
Hi @scottchiefbaker. Anything I can still do in this PR for you to reconsider it? I'm refactoring some of the current code here, and at the end the change only has 10 extra lines (and 5 of those are comment lines).

scottchiefbaker:
@ericbn thanks for bringing this back up... It slipped off my radar.

Ultimately I think it boils down to "do our users want this". It adds two additional lines to each commit line, and changes how the added/modified/removed lines work. Anytime there is change, users notice (and complain). I think it boils down to "is this a good enough addition to warrant the change", and I don't have a good answer to that.

Perhaps we should take a vote of the core developers to see what their take is?

vraravam:
I for one would definitely vote for this feature enhancement!

Svalorzen:
Any news about this change? This PR seems to have been here a while. I also support making more visible the distinction between different commits; after all, if one is trying to show multiple commits rather than one large patch it means that looking at each commit's diff individually is valuable, and so the start and end point of each commit should be more visible (certainly more visible than the header for each individual file change).

OJFord:
@scottchiefbaker wrote:
> Personally I don't think I would use, but I can definitely some people in the community wanting it. The issue then becomes that we need to make this a configuration option to allow people to opt in/out of this feature.

and later:
> Anytime there is change, users notice (and complain). I think it boils down to "is this a good enough addition to warrant the change", and I don't have a good answer to that.
> 
> Perhaps we should take a vote of the core developers to see what their take is?

personally I _do_ like it, but given a config option it could just be default off if concerned about that? Leave default-on for a major version bump?

scottchiefbaker:
I'm going to close this issue and move discussion over to #497.

PR #403, opened 2021-05-04
Fix MacOS Tests

### LINKED ISSUE #402: Need to fix failing OSX tests
Looks like our OSX tests are failing with some tooling errors. I don't know anything about OSX testing/tooling I'm hoping someone in the community can help.

GenieTim:
Fixes #402 

The implementation here is just an update of the MacOS version used by Travis, that seems to fix the issue #402, probably because the other one was quite outdated. I do not know what versions you want to be compatible with — this solution here is the latest MacOS (the first attempt I tried), but I can run some matrix to find the lowest version where the tests do not fail either (or look for other solutions).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,9 +15,9 @@ perl:
 matrix:
   include:
     - os: osx
-      osx_image: xcode9.4
+      osx_image: xcode12.2
       language: generic
-      perl: 5.24
+      perl: 5.30
   allow_failures:
     - perl: "blead"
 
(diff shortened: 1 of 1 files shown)


scottchiefbaker:
We have pretty good test coverage on Linux. I don't think we need to go nuts on OSX stuff (unless you really want to). I'm more concerned with "does it work on a semi-recent version of OSX and Perl". 

This looks like it tests against Perl 5.30 and some version of XCode. Forgive my ignorance but what is XCode and how does it relate the OSX?

GenieTim:
XCode is the IDE by Apple, bundled with all the build tools needed to build for an Apple device. In case of Travis, they use it as a measure of the MacOS Version to use.

PR #404, opened 2021-05-13
Link installation docs to known packages

### LINKED ISSUE #380: How to installon debian
diff-so-fancy doesn't appear in my repos. How do you manually install it?
`あす sudo aptitude install diff-so-fancy                                                                                     
[sudo] password for kenzo: 
Couldn't find any package whose name or description matched "diff-so-fancy"
Unable to apply some actions, aborting
`

OJFord:
Closes #380.

Is that wording appropriate @aos?  (I'm not a Debian user) I take it it is only available from your repo/'ppa' that has to be manually added, it's not just a case of `apt install` on a fresh system?

(Targetting master rather than next since it's a readme-only change relevant for existing versions, and that's the default displayed branch on GitHub.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,9 @@ Vanilla `git diff` vs `git` and `diff-so-fancy`
 
 Installation is as simple as cloning this repo and then putting the `diff-so-fancy` script in to your `$PATH`. The `lib/` directory will need to be kept relative to the core script.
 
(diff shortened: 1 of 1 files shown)


rootwork:
Thanks for updating the documentation with this -- yes, I think it helps!

stevemao (approved):

PR #406, opened 2021-05-18
upgrade to modern bats

paulirish:
(this is to `next` branch)

after we adopted **bats**, [it was forked](https://github.com/bats-core/bats-core#background) as the original repo stalled out. it's now maintained by the community so we can update to the latest core/assert/support components.  

I also **reverted** your 29839a0 "Fix the tests by using the previous ANSI defaults" because i preferred what you had. so instead, i changed the assertions a tad so it'd play well. 

lastly i updated all the **CI scripts** to use the new bats.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,16 +5,15 @@ jobs:
     steps:
       - checkout
       - run:
-          name:  
+          name:
           command: |
             sudo apt-get -y update
             sudo apt-get -y install shellcheck git
-            
-      - run: git clone --depth 1 https://github.com/sstephenson/bats.git 
+
+      - run: git submodule sync
       - run: git submodule update --init
       - run:
           name: tests
           command: |
-            export TERM=dumb && bats/bin/bats test/*.bats 
+            export TERM=dumb && ./test/bats/bin/bats test
             shellcheck *.sh
-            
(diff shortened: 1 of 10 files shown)


scottchiefbaker:
Does `update-deps.sh` work to pull in the new bats stuff?

paulirish:
> Does `update-deps.sh` work to pull in the new bats stuff?

hmm. well.. it does now. https://github.com/so-fancy/diff-so-fancy/pull/406/commits/471c1b528220a98a2cc007aecbfbdd147b772b70 :) 


btw, I did encounter a little weirdness with the bats-support submodule as the git repo moved, but our submodule is in the same path. for you and anyone with an existing clone... you'll definitely have to `git submodule sync; git submodule update --init` in the project root  (but it's possible you'll have more problems. if so.. i think it'd probably work out after `rm -rf test/test_helper/bats-*; git reset --hard; git submodule sync && git submodule update --init;` or something like that?? ) gl.

scottchiefbaker:
I confirmed that a clean:

```
git clone https://github.com/so-fancy/diff-so-fancy/
cd diff-so-fancy
git checkout next-bats
./update-deps.sh
bats test
```

Works... so we're good to go.

PR #407, opened 2021-05-18
run CI in github actions

paulirish:
I got CI w/ github actions working. It runs linux, mac and windows. and I have it testing all the perl versions we had in travis (except `blead` and `dev`).

linux certainly needed the same TERM=dumb hack we used elsewhere.

Generally I'm super happy with this and think this will allow us to stop using travis/ci/appveyor.

<details>
<summary>But.. there are 3 failures in Windows right now. </summary>
It seems that the test suite passes file in Cygwin bash (our appveyor env), but fails in Git Bash, which is what's used in the Github Actions environment. I could reproduce the same failures on my local windows machine. It looks like there may be some additional, unexpected linebreaks in the windows case. I can file more details about that.

filed in #408 
</details>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,65 @@
+name: ci
+
+on:
+  push:
+   branches: [master]
+  pull_request: # run on all PRs, not just PRs to a particular branch
+
+jobs:
+  unit:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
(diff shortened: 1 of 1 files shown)


webstech:
Does this still fail?  I just ran okay on `git bash`.  My branch is based on `next`.

scottchiefbaker:
@paulirish what is the status on this?

scottchiefbaker:
Closing for staleness. If this can be re-worked I would be willing to merge it.

PR #420, opened 2021-10-21
test: use isolated, temporary `GIT_CONFIG`

rwe:
I noticed that the unit tests were changing the configuration in `.git/config`. They also read user configuration as fallback values, which made things verifying correct behaviour confusing—especially when some user configuration matches the defaults!

(Although there was a specific guard checking against `in_unit_test`, that was only for boolean config values, not configuration in general).

This generates a gitconfig in `$BATS_TMPDIR` and exports both `GIT_CONFIG=…` and `GIT_CONFIG_NOSYSTEM=1`, which `git` will respect for isolating that configuration.

That setup and teardown is done in `setup_file` (once per suite, rather than once per test). Some other test setup has been moved into those bat callbacks as well.

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -490,24 +490,10 @@ sub git_config_boolean {
 	my $search_key    = lc($_[0] || "");
 	my $default_value = lc($_[1] || 0); # Default to false
 
-	# If we're in a unit test, use the default (don't read the users config)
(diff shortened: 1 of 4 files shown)


rwe:
I'm not seeing the failure, do you have a link?

PR #421, opened 2021-10-21
fix: over-permissive regex for added/deleted file meta header

rwe:
Noticed this because the line itself was filtered out of diffs in this codebase.

The previous pattern uselessly included `.*`, unlike all the other patterns, and so would erroneously match on actual diff content lines.

This also tightens the pattern since the file mode will always be six octal (0-7) chars; and in the git codebase the format string is: `"%s%snew file mode %06o%s\n"`.

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -303,7 +303,7 @@ sub do_dsf_stuff {
 		###################################
 		# Remove any new file permissions #
 		###################################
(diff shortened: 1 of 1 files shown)


scottchiefbaker:
This PR also looks good. I just noticed that last PR was against `master`. 

Per our [contribution guidelines](https://github.com/so-fancy/diff-so-fancy#contributing), all PRs should be against `next`. If you redo this against `next` I'll merge it.

PR #425, opened 2021-11-02
Re-enable less termcap initialization

codeinred:
This commit removes usage of the the -X option from diff-so-fancy, which prevented less termcap initialization. From the manpage:

> ```
> -X or --no-init
>     Disables sending the termcap initialization and deinitialization
>     strings to the terminal. This is sometimes desirable if the
>     deinitialization string does something unnecessary, like clearing
>     the screen.
> ```

The default behavior of `less` is to signal termcap initialization. This behavior provides a number of nice features, including that the output disappears when `less` is closed, and that mouse scrolling of the less viewport works automatically on terminals such as gnome-terminal.

I believe that this is a sensible default, and it ensures that diff-so-fancy works out of the box with fetaures users have come to expect of `less`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Issues relating to packaging ('installation does not work', 'version is out of d
 Configure git to use `diff-so-fancy` for all diff output:
 
 ```shell
-git config --global core.pager "diff-so-fancy | less --tabs=4 -RFX"
+git config --global core.pager "diff-so-fancy | less --tabs=4 -RF"
 git config --global interactive.diffFilter "diff-so-fancy --patch"
 ```
 
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -726,7 +726,7 @@ diff-so-fancy --set-defaults             # Configure git-diff to use diff-so-fan
 diff-so-fancy --patch                    # Use diff-so-fancy in patch mode (interoperable with `git add --patch`)
 
 # Configure git to use d-s-f for *all* diff operations
-git config --global core.pager \"diff-so-fancy | less --tabs=4 -RFX\"
+git config --global core.pager \"diff-so-fancy | less --tabs=4 -RF\"
 
 # Configure git to use d-s-f for `git add --patch`
 git config --global interactive.diffFilter \"diff-so-fancy --patch\"\n";
@@ -803,7 +803,7 @@ sub check_first_run {
 
 sub set_defaults {
 	my $color_config = get_default_colors();
-	my $git_config   = 'git config --global core.pager "diff-so-fancy | less --tabs=4 -RFX"';
+	my $git_config   = 'git config --global core.pager "diff-so-fancy | less --tabs=4 -RF"';
 	my $first_cmd    = 'git config --global diff-so-fancy.first-run false';
 
 	my @cmds = split(/\n/,$color_config);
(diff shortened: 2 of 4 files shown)


scottchiefbaker:
I'm not sure how I feel about this. `-RFX` has been the recommendation for several years now, and it hasn't caused any problems (bug reports). I've used it on my own d-s-f install and haven't had an issue. 

If it works for you that's great. I'm just not sure I'm ready to change the defaults/documentation unless we have a specific problem we're trying to solve.

scottchiefbaker:
Just so it's written down:

```
	-R or --RAW-CONTROL-CHARS
              Like -r, but only ANSI "color" escape sequences and OSC 8  hyperlink  se‐
              quences  are  output  in "raw" form.  Unlike -r, the screen appearance is
              maintained correctly, provided that there are no escape sequences in  the
              file  other than these types of escape sequences.  Color escape sequences
              are only supported when the color is changed within one line, not  across
              lines.   In other words, the beginning of each line is assumed to be nor‐
              mal (non-colored), regardless of any escape sequences in previous  lines.
              For  the  purpose of keeping track of screen appearance, these escape se‐
              quences are assumed to not move the cursor.
              
	-F or --quit-if-one-screen
              Causes less to automatically exit if the entire file can be displayed  on
              the first screen.
              
	-X or --no-init
              Disables sending the termcap initialization and deinitialization  strings
              to  the  terminal.   This  is sometimes desirable if the deinitialization
              string does something unnecessary, like clearing the screen.
```

At minimum I think `-R` is required. `-F` is a nicety, and depending on your terminal `-X` may not be required.

scottchiefbaker:
Did we come to a conclusion on this? I'm of the mindset, if it ain't broke don't fix it. If we want to document this behavior I'm fine with that, but I don't know if we need to change defaults.

AlanGabbianelli:
> FWIW, it has come up before: #26, #76, #246.
> 
> As I understand it the behaviour's entirely terminal dependent anyway? In mine (alacritty) it makes no difference to clearing or scrolling (nor anything else I can immediately notice) whether I use `-X` or not.
> 
> Personally I'd lean toward saying/setting nothing more than `| less` (or `| ${PAGER:-less}`!), any options you need/want on your system are up to you; maybe you've already set `$LESS`, etc.

@OJFord That's what I do.

I have this in my `~/.gitconfig`:
```
[core]
  # https://github.com/so-fancy/diff-so-fancy#with-git says to use less with the -RFX flags
  # but that prevents the diff to be mouse/trackpad scrollable. I simply use lees here
  # because I already use less with the -RF flags by default. See ~/.zshrc.
  # See https://github.com/so-fancy/diff-so-fancy/pull/425
  pager = diff-so-fancy | less --tabs=4
```

And this in my `~/.zshrc`:
```
# Unlike more, if the file content fits the screen, less will still display the prompt.
# To override this behaviour, specify the -F option. This causes less to automatically
# exit if the entire file can be displayed  on the first screen.
export LESS="$LESS -RF"
```

> Did we come to a conclusion on this? I'm of the mindset, if it ain't broke don't fix it. If we want to document this behavior I'm fine with that, but I don't know if we need to change defaults.

@scottchiefbaker Maybe this can be left as is, and documented? Even just something simple like linking to this PR or the issues mentioned above would suffice.

scottchiefbaker:
@AlanGabbianelli I think a documentation update is the best solution for this. Where do you think we should put it? Do you want to write something up in a PR?

scottchiefbaker:
I think this is better as a documentation fix only.

PR #434, opened 2022-04-15
Add a recipe to build snap packages

barryprice:
While I see there's already a PPA for Linux `.deb` packages, a `snap` package is a little easier on the end user (no need to add an external repository), and brings in automatic updates and more fine grained security controls.


barryprice:
FYI this is building snaps here, happy to hand that namespace over too if there's any interest in merging this:

https://snapcraft.io/diff-so-fancy

scottchiefbaker:
This is a decent idea... but I don't think it should go in `snap/`... 

Might make sense to put in `third_party/` somewhere? Maybe we make a buildsys dir? I imagine we might need the same thing for `.rpm` or `.deb` etc? I'm open to suggestions.

For the purposes of packaging up d-s-f the only thing you need is the fatpacked version that's attacked to each release. I tried to make d-s-f as simple as possible, and only require one file (no deps).

barryprice:
All good points, thanks - will refactor and get back to you!

barryprice:
So the issue with snap layout is that the build recipe *must* go in either /snapcraft.yaml or /snap/snapcraft.yaml (or .snapcraft.yaml):

https://github.com/canonical-web-and-design/build.snapcraft.io/issues/1094

I wonder if a better approach might be a separate repo, which seems to be how the traditional package is built, here:

https://github.com/aos/dsf-debian

I've set one up here to snap the fatpacked version as suggested:

https://github.com/barryprice/snap-dsf

Closing this out on that basis, but I'd be happy to hand the above over - or to just maintain it myself if that's not stepping on toes.

scottchiefbaker:
If you're confortable maintaining it, then by all means feel free. I don't know enough about that process, so I defer to someone more qualified.

Issue #394, opened 2021-01-02
Support for recursive flag


rightaway:
If using `diff --recursive` only the first result has the proper formatting of the file names `Renamed: file1 to file2` with the coloring etc. Would be good to see it on all files and have full support of recursive flag, including when a file is in one directory and not the other.

scottchiefbaker:
You're talking about vanilla diff? Not `git diff` output?

rightaway:
@scottchiefbaker Yes for both vanilla diff and git diff, I'd like to pass it to diff-so-fancy even if it's `--recursive` and have formatted result.

scottchiefbaker:
Apologies, this slipped through the cracks. If you can provide more information I might be able to tackle this. I'm not familiar with `diff --recursive`. How do you use it in your workflow? Can you provide a simple diff that show d-s-f providing the wrong output?

scottchiefbaker:
Here is a possible simple test case (I think).

```diff
Only in /var/tmp/b: donk
diff --recursive -u /var/tmp/a/foo/bar /var/tmp/b/foo/bar
--- /var/tmp/a/foo/bar  2026-03-13 21:35:28.492228513 -0700
+++ /var/tmp/b/foo/bar  2026-03-13 21:35:47.101220189 -0700
@@ -1 +1 @@
-BAR
+FOO
diff --recursive -u /var/tmp/a/index.txt /var/tmp/b/index.txt
--- /var/tmp/a/index.txt        2026-03-13 21:35:20.997231861 -0700
+++ /var/tmp/b/index.txt        2026-03-13 21:35:41.030222906 -0700
@@ -1 +1,2 @@
 INDEX
+INDEX2
```

scottchiefbaker:
This should be fixed in the latest v1.4.6 release. If not, please re-open.

PR #436, opened 2022-06-09
Support vanilla diff recursive flag

brewsfab:
Thank you so much for the great project you brought to us.

Needed support for the recursive flag so I made this improvement thinking it could help more people.
It basically treats the vanilla diff with recursive flag as Mercurial.

- Related to #394 
- All tests are passing.


scottchiefbaker:
I consider myself fairly well versed in regular expressions, but man what you wrote is beyond me :)

I'd like the comment above the regexp adjusted so the `$4` and `$5` stuff line up with the appropriate parens, and then a comment added explaining the logic of the regexp so when someone has to look at it in the future we'll have a starting point.

scottchiefbaker:
Otherwise things look good. Clean diff and solid unit tests.

brewsfab:
Sorry for missing that part about the patterns positions.
Since the $4 is very long I have decided to add some decoration to show its scope.
I also had some comments about the regex, I hope it is easy to understand.
Let me know!

scottchiefbaker:
This looks good, and a lot clearer to me :)

What string are we actually matching against in a recursive diff. Do you have an example?

scottchiefbaker:
Since all we're trying to do is capture `-r` or `--recursive` somewhere after the `diff` wouldn't it be easier to just do:

```
(\b-r\b)|(--recursive)|(--git)|(--cc))
```

That will ensure the `-r` is not part of another word and not a double `--`, and would be more readable?

brewsfab:
Sorry I was not clear in the answer. My bad.
In fact, all the variants below are valid for recursive diff.

1. diff -ur folder1/filer1 folder2/filer1
2. diff -u -r folder1/filer1 folder2/filer1
3. diff --recursive folder1/filer1 folder2/filer1
4. diff --color -u -r folder1/filer1 folder2/filer1
5. diff --color -r -u folder1/filer1 folder2/filer1
6. diff --color --recursive folder1/filer1 folder2/filer1
7. diff -bur folder1/filer1 folder2/filer1
8. diff -rub folder1/filer1 folder2/filer1
9. diff -r -ub folder1/filer1 folder2/filer1
10. diff -bru folder1/filer1 folder2/filer1

So the current proposal covers all 10 variants. [regex101.com complex sample](https://regex101.com/r/OJYTS0/1)
Whereas the more readable 
```
diff (-r\b|--recursive|--git|--cc) (.*?)(\e| b\/|$)
``` 
will cover only the number 3 and number 9 variants. [regex101.com readable sample](https://regex101.com/r/iyjPUm/1)

**I think there should be a trade-off between readability and adaptability.**

Since it is user managed more than the *git diff* which is somehow standardized as `diff --git` at the beginning of the line, there might be different flavors for the diff recursive.
I guess there are people like me with alias configuration as:
```sh
alias diffr='diff -br' #diff recursive ignoring whitespace changes
```
or much more complex.
So for us, doing `diffr -u folder1 folder2 | diff-so-fancy` will fail to use the power of diff-so-fancy. 

Also forgetting to remember to use the flag -r separated made me re-run the diff a second time every time to fix it.

So, if the choice is the readability one option to consider is to enforce/clarify in the docs to use -r (separated) or --recursive flags as first arguments.
So people will change their default configuration for diff recursive to something like
```sh
alias diffr='diff -r -b <extra_args>' #diff recursive ignoring whitespace changes
```
or
```sh
alias diffr='diff --recursive -b <extra_args>' #diff recursive ignoring whitespace changes
```

Let me know about your choice so I will make change to the PR to suit if needed.

scottchiefbaker:
@brewsfab to hit on everything I was thinking. No need to over-complicate the code a simple feature. I think we should have support for two calling methods:

```
diff -r -u file_a file_b | diff-so-fancy
or
diff --recursive -u file_a file_b | diff-so-fancy
```

and update the documentation to reflect that the recursive flag needs to be immediately after the diff like you mentioned. I think that is a good compromise if simplicity and feature support. If you want to update this pull request to reflect that I will land it.

brewsfab:
Thanks.

I have simplified the regex and added the comment as requested.
Please verify.

scottchiefbaker:
Looks great... thanks for all the hard work and the revisions!

Kumm-Kai:
It would be great if the instructions from the readme would also be added to the `--help` output.

scottchiefbaker:
Pull requests welcome

PR #441, opened 2022-09-12
Add a reset to remove bold when not asked

utix:
Some part are bold even if not configure to be

`func` on this example:

image

Reseting ascii at the begining fixes it

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -300,7 +300,7 @@ sub do_dsf_stuff {
 				print $frag_color;
 			}
 
-			print "@ $last_file_seen:$start_line \@${bold}${last_function_color}${remain}${reset_color}\n";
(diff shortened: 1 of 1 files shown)


scottchiefbaker:
Is the only place this is affecting the "last function" fragment? If so, it probably makes more sense to JUST reset that one, instead of EVERYTHING in `git_ansi_color()`?

utix:
Hard to check all combination, should I check on master or next branch ?
For info bright, seems broken on next branch, since  754f9dfbd2085eb8241d71808dda9592b463c5a8 with git bisec
for example birghtyellow is not working on func

utix:
Ok I get it, https://github.com/so-fancy/diff-so-fancy/blob/master/diff-so-fancy#L303 bold seems forced here

scottchiefbaker:
This looks innocent enough, and seems like a good solution.

PR #442, opened 2022-09-12
Add dim and underline as Select Graphic Rendition

utix:
first frag with dim
second func with ul

image

configs:
```
[color "diff"]
        meta = 11
        frag = yellow dim
        old = red bold
        new = green bold
        func = brightyellow
        commit = yellow bold
        whitespace = red reverse
```

```
[color "diff"]
        meta = 11
        frag = yellow
        old = red bold
        new = green bold
        func = brightyellow ul
        commit = yellow bold
        whitespace = red reverse
```

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -907,6 +907,14 @@ sub git_ansi_color {
 		push(@ansi_part, "1");
 	}
 
+	if (grep { /dim/ } @parts) {
+		push(@ansi_part, "2");
+	}
+
+	if (grep { /ul/ } @parts) {
+		push(@ansi_part, "4");
+	}
+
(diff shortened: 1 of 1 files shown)


scottchiefbaker:
The lines that remove the ul/dim from the array aren't necessary because it's handled right after the code for reverse. Remove those two lines and I'll land this.

scottchiefbaker:
Repurposing my comment from another thread so it's documented in both places:

>Just an FYI we purposely did not implement all the Git specific color wordings. There are too many, and the idiosyncrasies were too much to bother with. I'm not against adding some of them in, but we need to be aware the demand for some of these corner cases has been about zero (until this PR).

This one looks simple enough though, so I'll land it.

PR #444, opened 2022-09-12
Generate bright colors

### LINKED ISSUE #443: Bright colors are not working when color is not coming from git
Since 754f9dfbd2085eb8241d71808dda9592b463c5a8 bright colors for func are not working

utix:
Closes #443 

:warning: bright and bold are different

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -900,6 +900,9 @@ sub git_ansi_color {
 		'cyan'    => 6,
 		'white'   => 7,
 	};
+	for my $k (keys %{ $colors }) {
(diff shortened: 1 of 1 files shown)


scottchiefbaker:
Just an FYI we **purposely** did not implement all the Git specific color wordings. There are too many, and the idiosyncrasies were too much to bother with. I'm not against adding some of them in, but we need to be aware the demand for some of these corner cases has been about zero (until this PR).

That said this one looks pretty simple, and actually removes more lines than it adds. It passes all the unit tests so I'm OK with this one. Good work!

Issue #461, opened 2023-03-05
`report-bug.sh` is broken? May need to be modernized


scottchiefbaker:
@stevemao it was reported in #460 that the `report-bug.sh` script is broken. When I run that on my server (via SSH) I get clipboard errors as well. Looks like we need to check **if** they have a clipboard application installed at all, and then if that command can even read the clipboard, which may not work on non-local machines.

Or maybe we need to rethink that whole script? I dunno.

With `xclip` installed I get: `Error: Can't open display: (null)`
With nothing installed I get `clipboard is unavailable`. 

We should probably offer some suggestions to the user about how to solve this problem?

kambala-decapitator:
please note that my error was not about clipboard:

```
> curl https://raw.githubusercontent.com/so-fancy/diff-so-fancy/master/report-bug.sh | bash -s 'git diff sigma-2.4 tbl/sigma/en/patchstring.txt' file.txt
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1262  100  1262    0     0   9600      0 --:--:-- --:--:-- --:--:-- 10015
fatal: option '--color' must come before non-option arguments
```

scottchiefbaker:
This should be addressed in 0b39dfb

PR #470, opened 2023-06-09
Possible fix for #461

scottchiefbaker:


scottchiefbaker:
Oh wow this is a bad merge between branches... Cleaning up first

PR #471, opened 2023-06-10
Add Fedora to known packages

principis:
The README seems to be overwritten due to the merge of the next branch. This PR adds the Fedora package again, but this time on the next branch. :slightly_smiling_face:

It would be nice if this could be merged to master before the next update, as it's preferable for users to use the official package.


scottchiefbaker:
Thank you for this... there was like 17x merge conflicts on this section when I was building the release. Clearly I didn't merge everything 100%.

PR #479, opened 2023-11-21
Update Arch Linux package URL in README.md

felixonmars:
The old URL returns 404 now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Vanilla `git diff` vs `git` and `diff-so-fancy`
 
 Installation is as simple as cloning this repo and then putting the `diff-so-fancy` script in to your `$PATH`. The `lib/` directory will need to be kept relative to the core script.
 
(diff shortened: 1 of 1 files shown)


OJFord:
Thanks!

OJFord (approved):

scottchiefbaker:
I have no problem with this commit and am happy to merge it. [Pull requests](https://github.com/so-fancy/diff-so-fancy#contributing) should target the `next` branch however. If you update this to hit that branch I'll merge it.

OJFord:
@scottchiefbaker FYI I've changed the default branch to `next` so that this stops happening. It's also necessary for this readme change to actually appear in the rendered readme when people navigate to the repo (and don't change the branch, because why would they) without also updating `master` too.

felixonmars:
I actually didn't know that I could change the target branch right in this PR. Thanks!

PR #482, opened 2024-02-24
Make this loadable easily by ZSH frameworks

unixorn:
Let `zsh` users with frameworks use the framework to automatically clone the repostitory and add it to their `$PATH` so that the frameworks can keep the `git` checkout up to date.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,8 @@ Vanilla `git diff` vs `git` and `diff-so-fancy`
 
 Installation is as simple as cloning this repo and then putting the `diff-so-fancy` script in to your `$PATH`. The `lib/` directory will need to be kept relative to the core script.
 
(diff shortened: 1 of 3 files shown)


scottchiefbaker:
I don't know a whole lot about `zsh` but I support this effort. 

Have you looked at the `zsh` stuff in the pro tips documentation? Is there any overlap there? If you want to move your stuff to the protips docs instead of the main README I'll land this code.

unixorn:
I'll update the doc there, then people can use dsf directly as a plugin without hassling with submodules. I think it's cleaner to have the framework support integrated in the tool repo.

unixorn:
Updated the doc per your suggestion & updated `diff-so-fancy.plugin.zsh` to conform better to the ZSH plugin standard.

unixorn:
Thanks!

PR #484, opened 2024-03-01
Fix broken link in README.md

### LINKED ISSUE #483: Broken link in ZSH install instructions
In https://github.com/so-fancy/diff-so-fancy/tree/578bffcc5a7278e708891b224d2637c382c238da?tab=readme-ov-file#install, the `README.md` points to `diff-so-fancy/zsh-frameworks.md`, which doesn't actually exist.

Can we update the ZSH install instructions to reflect https://github.com/so-fancy/diff-so-fancy/pull/482?

Cc @unixorn

unixorn:
Closes #483

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Vanilla `git diff` vs `git` and `diff-so-fancy`
 
 Installation is as simple as cloning this repo and then putting the `diff-so-fancy` script in to your `$PATH`. The `lib/` directory will need to be kept relative to the core script.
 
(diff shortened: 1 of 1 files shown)


scottchiefbaker:
The link should target `/blob/next/pro-tips.md` instead of a full path. Otherwise it looks fine. Update the target and I'll land this.

unixorn:
Done!

scottchiefbaker:
Sorry I wasn't clear... You should drop the `http://...` part of the link. That will make the link work on alternate branches and not always link to `main`. Just linking to `/blob/next/pro-tips.md#zsh-plugin-suppport-for-diff-so-fancy` should be sufficient.

unixorn:
No problem, done.

PR #486, opened 2024-03-01
Add awesomebot link checker

unixorn:
Add a github action to check the links in README.md

This will find broken links like #483 during PR checks

### DIFF
--- a/.github/workflows/awesomebot.yml
+++ b/.github/workflows/awesomebot.yml
@@ -1,0 +1,23 @@
+---
+name: Check links in README
+
+on:
+  push:
+    branches: [ '*' ]
+  pull_request:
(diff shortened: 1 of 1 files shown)


scottchiefbaker:
Can you elaborate a little bit? What exactly will this do?

Check the links how?

unixorn:
For each link, it will make sure that when it tries to reach it, there are no errors.

Basically it does a liveness test. The configuration I'm adding allows redirects to not be errors and ignores some of the temporary 50x server errors.

unixorn:
I use it to keep the [awesome-zsh-plugins](https://github.com/unixorn/awesome-zsh-plugins) list from developing link rot.

PR #497, opened 2024-10-18
Draw ruler and box around headers - rebased

Svalorzen:
This PR is the same as PR #398, I have just rebased it onto the current next so that there are no conflict. Just some re-tabbings were needed.

I'm doing this as the PR still seems to work well, and it would be a shame to lose the amount of work that @ericbn did.

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -28,11 +28,37 @@ my $git_strip_prefix           = git_config_boolean("diff.noprefix","false");
 my $has_stdin                  = has_stdin();
 my $CONTEXT_LINES              = undef; # Number of lines of context diff used
 
+my $box_horizontal;
+my $box_vertical;
+my $box_down;
+my $box_up;
+# BOX DRAWINGS LIGHT HORIZONTAL http://www.fileformat.info/info/unicode/char/2500/index.htm
+# BOX DRAWINGS LIGHT VERTICAL https://www.fileformat.info/info/unicode/char/2502/index.htm
+# BOX DRAWINGS LIGHT DOWN AND LEFT https://www.fileformat.info/info/unicode/char/2510/index.htm
+# BOX DRAWINGS LIGHT UP AND LEFT https://www.fileformat.info/info/unicode/char/2518/index.htm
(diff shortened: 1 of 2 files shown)


scottchiefbaker:
Couple of quick questions while I review this:

1. Can you provide a screenshot of what the output before/after this change?
2. Does this pass all the current unit tests?
3. This is a pretty large change so this feature would need to be behind `git config` option so users can **opt-in** to it

OJFord:
@scottchiefbaker fwiw there's a comparison screenshot in the original PR: https://github.com/so-fancy/diff-so-fancy/pull/398#issuecomment-782520586

Svalorzen:
I have to admit I'm not really familiar with the codebase, nor I really know Perl. This is mostly a feature that I have wanted for a long time, and after seeing the older PR stay there for a while I just thought I'd give it a spin, and I didn't really do a lot of work to make it run again -- all credit goes to the original author.

AppVeyor seems to say that the build is failing, so I'm assuming some tests do fail; perhaps looking for full lines where partial lines get outputted with this patch. I think those tests did not exist when the patch was first proposed.

If there's anything I can do with my limited knowledge to help this be put in, I can try, but as I said I'm unfamiliar with both the language and the codebase so I'm limited in what I can offer.

Svalorzen:
In fact, I can already attest to a minor bug: if a file changed takes more than one line, the line wraps are incorrect.

Current:
image

With PR:
image

scottchiefbaker:
Apologies for the delay on this. If this can be cleaned up against the code on `next` I will definitely consider merging it. Or perhaps we need to reopen discussion on exactly what we want out of this feature first?

What I'm seeing in previous discussion and screenshots is that 95% of these changes would only affect raw `diff -u` output, not `git diff` output? If we're talking about grouping by commit etc that's already done in git mode, but not raw diff mode.

scottchiefbaker:
I just landed some code in b32a4ef989bc47711e6de54b5edc4ace887a2991 that should address some of this. Please check the latest code in `next` and let me know if it meets your needs.

Issue #500, opened 2025-02-02
Change istructions to include necessary other files.


fNBU:
Currently the instructions 

https://github.com/so-fancy/diff-so-fancy/blob/36dcfde9dd82cd3b96abd4e82ecf8bee796381ff/README.md?plain=1#L13

are a bit confusing. The first sentence says explicitly that only the file `diff-so-fancy` needs to be copied (which is not correct). The second sentence implies (but does not explicitly say) that the directory `lib` needs to also be copied to the same location.

This has been mentioned before in eg https://github.com/so-fancy/diff-so-fancy/issues/318. Regardless of the packaging of a "fatpack" discussed there, the instructions should conform to whatever packaging is used.

fNBU:
PR: https://github.com/so-fancy/diff-so-fancy/pull/501

scottchiefbaker:
I will update the wording to say "Github Release" since that's the correct method for using this (unless you're develing)

PR #501, opened 2025-02-02
#500: fix instructions to include needed files

fNBU:
Adjusted wording to fix ambiguity

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Vanilla `git diff` vs `git` and `diff-so-fancy`
 
 ## Install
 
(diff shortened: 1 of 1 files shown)


scottchiefbaker:
The instructions reference "the script from the latest release" which is the official [GitHub releases](https://github.com/so-fancy/diff-so-fancy/releases) in the assets section. These are the pre-packaged all-in-one versions of the script with the library fatpacked in. The official GH releases do **not** require the `lib/` directory.

If you'd like to work on or test the build from the repo you need the d-s-f script and the `lib/` directory in the same directory in your path.

PR #505, opened 2025-05-06
Remove duplicate line in README

Ben3eeE:
Noticed this when reading through the README. From the history it seems like the double quoted version is the most recent and the other one should be removed :)


scottchiefbaker:
Weird... good catch.

PR #508, opened 2025-08-29
Update DiffHighlight.pm to latest version

IsaacOscar:
This updates the included `DiffHighlight.pm` to the [latest one](https://github.com/git/git/blob/702d8c1f3b5377a64670b0f22add157b0bfc33dd/contrib/diff-highlight/DiffHighlight.pm)

In particular, this is needed for `italic` in `color.diff-highlight.oldhighlight`/`color.diff-highlight.newhighlight` to work correctly.

### DIFF
--- a/lib/DiffHighlight.pm
+++ b/lib/DiffHighlight.pm
@@ -1,6 +1,6 @@
 package DiffHighlight;
 
-use 5.008;
+require v5.26;
 use warnings FATAL => 'all';
 use strict;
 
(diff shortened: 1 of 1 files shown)


scottchiefbaker:
We have a modified `DiffHighlight.pm`, hence all the "noise" in this PR. Nothing in here seems relevant to "italic" except maybe `my $s = `git config --get-color $key 2>$NULL`;`

Can you test with ONLY that line and see if you get what you need? We shouldn't need any of the other changes.

scottchiefbaker:
Actually... it's pretty minor I went ahead and cherry picked the parts we need in 3132702b6b0d0beaa6f9846e14b2a78731468f30.

PR #509, opened 2026-03-14
Update awesomebot and add dependabot

### LINKED ISSUE #485: Add awesomebot link checks on README.md
Awesomebot check would have caught #483 during PR checks

unixorn:
Here's my awesomebot configuration from the awesome-zsh-plugins repo. I left in a commented out command that has all my whitelisted sites as an example. I don't think you need them all, so as is, this just whitelists some of the badge sites.

I also added a dependabot configuration so that it will help keep your repo up to date.

This should close https://github.com/so-fancy/diff-so-fancy/issues/485 for you.

Thanks for diff-so-fancy, I use it literally every day.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+---
+# Use `allow` to specify which dependencies to maintain
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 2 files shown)


scottchiefbaker:
I don't know what a lot of this means, but I'll do anything to get Awesomebot to stop complaining for EVER commit I land :)

Really appreciate you working on this

unixorn:
No problem - I realized it was easier for both of us if I cloned in my working configs.

PR #510, opened 2026-03-19
add proper guide for publishing to npm

paulirish:


scottchiefbaker:
I only understand about 10% of this, but I trust you so good-to-go :)

PR #520, opened 2026-03-28
Fix spelling mistake in code comment

jgarber623:
As straightforward as it gets! Changed `diretory` to `directory`.

Also, looks like GitHub's online editor added a trailing newline. 🤷🏻


scottchiefbaker:
Thank you for your dilligence

Issue #511, opened 2026-03-19
Setting diff-so-fancy as `core.pager` falsifies git-grep results


AndiDog:
Easy to reproduce:

```
cat > /tmp/gitconfig.failing <<EOF
[core]
	pager = diff-so-fancy | less --tabs=4 -RF
EOF

mkdir /tmp/repo
cd /tmp/repo
git init
echo mystring > myfile
git add myfile

# This prints nothing
GIT_CONFIG_GLOBAL=/tmp/gitconfig.failing git grep mystring

# This prints the expected 'myfile:mystring'
GIT_CONFIG_GLOBAL=/dev/null git grep mystring
```

This was really surprising – if you have 2 lines containing `mystring`, not just 1, it works.

My workaround is this:

```
[core]
	pager = less --tabs=4 -RF
[pager]
	diff = diff-so-fancy | less --tabs=4 -RF
```

Maybe the README can be adapted to recommend the above? See documentation for [`pager.<cmd>`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-pagercmd).

scottchiefbaker:
I would be open to a pull-request to update the documentation for this. This is pretty specialized so I think it should go in `pro-tips.md`

flohw:
Hi,
It may be related to my issue: I have an alias `glol='git log --graph --pretty='\''%Cred%h%Creset -%C(auto)%d%Creset %s %Cgreen(%cr) %C(bold blue)<%an>%Creset'\'`. 

With diff so fancy 1.4.4, using `glol -1` showed me the latest commit. Now (1.4.6, where is 1.4.5 BTW? ^^) it prints nothing but exit successfully.

`core.pager=pager = diff-so-fancy | less --tabs=4 -RFX`

Let me know if you need another issue.

scottchiefbaker:
This should be a separate issue as they're unrelated. I'm confused though. You're using d-s-f to consume the `git log` output? d-s-f is only targeted at diff output so I'm not sure what to even expect when viewing the `git log` output filtered through d-s-f.

flohw:
I am not entirely sure why there is this behavior after the update. I don't use diff-so-fancy directly for git log, only indirectly for interactive.diffFilter but the standard `git log -1 -p` looks normal.

[...]

PR #524, opened 2026-04-11
Narrow diff-so-fancy pager recommendations

booxter:
Sometimes `diff-so-fancy` swallows input from `git grep`. This can be
mitigated by setting the pager for specific commands where we should
expect diff output.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,12 +20,23 @@ Issues relating to packaging ("installation does not work", "version is out of d
 
 ### Git
 
-Configure git to use `diff-so-fancy` for all diff output:
+Configure git to use `diff-so-fancy` for common diff output:
 
 ```shell
-git config --global core.pager "diff-so-fancy | less --tabs=4 -RF"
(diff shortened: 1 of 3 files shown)


scottchiefbaker:
Will you please open an issue to discuss this with the larger group. Conceptually it's good, but this touches a lot of files. d-s-f has been public for many years now, and this is the first time it's come up. Perhaps there is a way to make d-s-f more `git grep` friendly.

booxter:
I did some digging. What I hit with `git grep` was https://github.com/so-fancy/diff-so-fancy/issues/511 (I was on 1.4.6). I bumped to 1.4.10 and it no longer happens. If the tool is supposed to work with `git grep` output and it's just a bug, then I guess this change is not necessary.

scottchiefbaker:
Glad it's fixed for you. I'd **like** for d-s-f to work with all `git` output, but that's a pretty lofty goal.

PR #527, opened 2026-04-20
Suppress Perl warning for 'use_color' variable in Dump::Krumo

### LINKED ISSUE #526: Perl warning: "Name used only once: possible typo" for `Dump::Krumo::use_color`
### Description

When running `diff-so-fancy`, Perl emits the following warning:

```
Name "Dump::Krumo::use_color" used only once: possible typo at diff-so-fancy line 1383.
```

### Root cause

In the developer debug block near line 1383, the package variable `$Dump::Krumo::use_color` is assigned but never read anywhere else in the script. Perl's `warnings` pragma flags any package variable that appears only once as a potential typo — regardless of whether the surrounding `if` block is actually entered at runtime.

```perl
# If you're me you load a debug library
if ($ENV{USER} && ($ENV{USER} eq "bakers")) {
    require Dump::Krumo;
    $Dump::Krumo::use_color = 2;  # <-- triggers "used only once" warning
    Dump::Krumo->import(qw(k kd));
}
```

### Expected behavior

No warnings are emitted during normal use.

### Fix

Wrap the assignment in a `no warnings 'once'` block to suppress the warning locally:

```perl
if ($ENV{USER} && ($ENV{USER} eq "bakers")) {
    require Dump::Krumo;
    { no warnings 'once'; $Dump::Krumo::use_color = 2; }
    Dump::Krumo->import(qw(k kd));
}
```

cgrothaus:
### Summary

Suppresses a spurious Perl warning emitted during normal use of `diff-so-fancy`:

```
Name "Dump::Krumo::use_color" used only once: possible typo at diff-so-fancy line 1383.
```

### What changed and why

The assignment `$Dump::Krumo::use_color = 2` inside the developer-only debug block is the only reference to that package variable in the entire script. Perl's `warnings` pragma flags this at compile time as a potential typo, even though the block is never entered for any user other than the author.

Wrapping the assignment in a `{ no warnings 'once'; ... }` block suppresses the warning locally and lexically, without affecting any other warning behaviour in the script.

### Testing

- Verified the warning no longer appears when running `diff-so-fancy` with `perl -w`.
- No functional behaviour is changed. The debug block is unaffected for the intended user.

Closes #526

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -1380,7 +1380,7 @@ BEGIN {
 	# If you're me you load a debug library
 	if ($ENV{USER} && ($ENV{USER} eq "bakers")) {
 		require Dump::Krumo;
-		$Dump::Krumo::use_color = 2;
+		{ no warnings 'once'; $Dump::Krumo::use_color = 2; }
 		Dump::Krumo->import(qw(k kd));
 	}
 }
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -1379,8 +1379,10 @@ BEGIN {
 
 	# If you're me you load a debug library
 	if ($ENV{USER} && ($ENV{USER} eq "bakers")) {
+		no warnings 'once';
+
 		require Dump::Krumo;
-		{ no warnings 'once'; $Dump::Krumo::use_color = 2; }
+		$Dump::Krumo::use_color = 2;
 		Dump::Krumo->import(qw(k kd));
 	}
 }


scottchiefbaker:
Good catch. No need to be quite so narrow with the scope. If you change this to scope it to the whole if I'll accept it.

```perl
    # If you're me you load a debug library
    if ($ENV{USER} && ($ENV{USER} eq "bakers")) {
        no warnings 'once';

        require Dump::Krumo;
        $Dump::Krumo::use_color = 2;
        Dump::Krumo->import(qw(k kd));
    }
```

cgrothaus:
@scottchiefbaker Thanks! I fixed the scope of the "no warning".

PR #535, opened 2026-06-18
Fix patch mode line mismatch on added/deleted files

arnaud-dezandee:
Using diff-so-fancy as an `interactive.diffFilter` (`git add -p`) currently fails with:
```
error: mismatched output from interactive.diffFilter
hint: Your filter must maintain a one-to-one correspondence
hint: between its input and output lines.
```
whenever a file is added or deleted with content. git requires the filter to emit exactly one output line per input line, and diff-so-fancy was dropping the newlines it uses to keep that 1:1 ratio.

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -164,12 +164,14 @@ sub do_dsf_stuff {
 	my $input     = shift();
 	my $chunk_str = join("", @$input); # Store the full chunk text for lookbacks
 
-	# FIXME: There is a scenario where there are FIVE diff header lines
-	# I'm not sure if we need to account for that or not. Unit tests pass so... bees?
+	# git requires interactive.diffFilter output to stay 1:1 with input. Add/delete
+	# with content have 3 header lines (not 4) but still need balancing; empty ones (no @@) don't.
 	my $cnt = count_git_header_lines(@$input);
-	if ($cnt == 4) {
+	if ($cnt == 4 || ($cnt == 3 && grep { /^${ansi_color_regex}@@/ } @$input)) {
 		$patch_mode    = 1;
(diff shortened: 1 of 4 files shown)


scottchiefbaker:
We did a TON of work to address all the corner cases for patch mode. Can you provide a simplified test case that reproduces this problem so I can investigate?

arnaud-dezandee:
> We did a TON of work to address all the corner cases for patch mode. Can you provide a simplified test case that reproduces this problem so I can investigate?

Thanks for taking a look! Here's a minimal repro. It has to be run in a real terminal
```
repo=$(mktemp -d) && cd "$repo"
git init -q
git commit -q --allow-empty -m init
git config interactive.diffFilter "diff-so-fancy --patch"

printf 'line one\nline two\nline three\n' > newfile.txt
git add -N newfile.txt   # intent-to-add so `git add -p` sees the new file

git add -p
```

scottchiefbaker:
Great reproducible test case. I can confirm this is a bug, and your code does fix it.

Minor fix needed before I will land this. For bugfixes I like the unit tests in one commit, and then the actual code fix in a second commit. This allows me to confirm the code is broken by running the unit tests, and then fixed after the code is applied. If you can break the PR apart into two commits in this fashion I will gladly land it.

arnaud-dezandee:
> If you can break the PR apart into two commits in this fashion

Done, first commit failed CI and second fixes it

scottchiefbaker:
1. PR looks good, I now see two commits
2. One test fails without the fix in place
3. All tests pass with the fix
4. Code looks minimal and easy to understand

Good work! I wish all my PR were this simple and well written.

Issue #540, opened 2026-07-16
Error from v1.4.12: `Use of uninitialized value $time in split at /opt/homebrew/bin/diff-so-fancy line 1525`


fannheyward:
```
[core]
    pager = diff-so-fancy | less --tabs=4 -RF
    excludesfile = ~/.config/git/ignore
    commitGraph = true
    preloadindex = true
    untrackedCache = true
[interactive]
    diffFilter = diff-so-fancy --patch
```

No such error with v1.4.10

scottchiefbaker:
That code should only trigger in `git show` mode, and only on a line that starts with `Date:`. Are you able to provide the raw git output so I can see where the code is failing?

fannheyward:
I came this error with `git show` and `git log` too. Here's a `git show` output:

```
commit 56bf841b1adffdad3c650e021462cfb9a438b42e (HEAD -> master, origin/master, origin/HEAD)
Author: Heyward Fann <fannheyward@gmail.com>
Date:   2026-07-16 10:28:26 +0800

    update

diff --git a/css/main.css b/css/main.css
index dc33621..02da18d 100644
--- a/css/main.css
+++ b/css/main.css
@@ -240,7 +240,6 @@ pre > code {
  */
 .wrapper:after,
 .footer-col-wrapper:after {
-  content: "";
   display: table;
   clear: both;
 }
(END)
```

My `Date` config:

```
[log]
	date = iso
```

fannheyward:
Update: Removing the `date = iso` configuration resolved the error. How to make `diff-so-fancy` work with `date = iso`?

scottchiefbaker:
Today I learned about `date = iso`

According to the docs there are five options: `relative`, `local`, `default`, `iso`, and `rfc`.

I just landed 27e0ce5a16f5805825298789c7ba002c487c241f which fix `--date iso` please test and let me know.

fannheyward:
@scottchiefbaker tested with HEAD commit, works as expected!

mitchblank:
I hope this triggers a release soon.  I just did a `brew upgrade` and tripped over this regression in the first few minutes :-(

PR #544, opened 2026-08-11
Fix crash on git dates without a timezone offset

gabrielfeo:
To reproduce:

```
git init repo && cd repo
git commit --allow-empty -m "test"
git -c log.date=local show --color=always | diff-so-fancy
```

**Expected:** the commit metadata prints normally, with a human readable "(N ago)" annotation appended to the `Date:` line.

**Actual:** diff-so-fancy dies right after printing the `Author:` line, before the `Date:` line is even reached:

```
Use of uninitialized value $tz in division (/) at diff-so-fancy line 1549, <STDIN> line 5.
```

Seems to have been introduced in [382ff82](https://github.com/so-fancy/diff-so-fancy/blob/382ff82f1efffb4d040e0b4daf37e85f445f4569/diff-so-fancy#L1533-L1536). Related to #540, which reported the a similar crash for `log.date=iso`.

Cause: `git_datestr_unixtime()` matches the date line's prefix with a regex, then destructures `$wday, $mon_str, $mday, $hour, $min, $sec, $year, $tz` from the split fields. `log.date=local` (and `--date=local`) omit the trailing timezone field, so the split returns one fewer element and `$tz` comes back undefined. That reaches `int($tz / 100)` a few lines later, and since the script runs with `use warnings FATAL => 'all'`, the "uninitialized value" warning is fatal instead of just noise.

Fix: default `$tz` to `0` when the date string has no timezone field.

### DIFF
--- a/diff-so-fancy
+++ b/diff-so-fancy
@@ -1545,6 +1545,11 @@ sub git_datestr_unixtime {
 		return "";
 	}
 
+	# git's --date=local (and log.date=local) omit the timezone offset, so
+	# treat that as an already-local time with a zero offset instead of
+	# dying on the undefined value below (warnings are fatal in this file).
+	$tz //= 0;
+
 	# Convert local time to UTC using timezone offset like -700
 	my $offset_hours   = int($tz / 100);
 	my $offset_minutes = abs($tz % 100);


scottchiefbaker:
I swear I tested all the `--date XYZ` options, but clearly I missed local. Fix looks simple and well written. Thank you.

PR #556, opened 2026-09-19
Take the rendering width from $COLUMNS when it is set

### LINKED ISSUE #555: Use the COLUMNS env var to learn the terminal width
Currently, diff-so-fancy queries only the terminal using `tput cols` (which makes the TIOCGWINSZ ioctl call) to know the window width, and falls back to a default of 80 when output is not a tty. This is a problem for hosts running diff-so-fancy in a pipe in order to render its output in a view.

Lazygit is such a host: it used to run the git command with diff-so-fancy as a GIT_PAGER in a pty, but we recently had to change this on Windows because our pty implementation there won't work well with the upcoming OSC 1717 feature. We now run git and diff-so-fancy in a pipe without a pty, and setting COLUMNS would allow us to not make this a breaking change for users.

Here's a PR that implements this: #556. Disclaimer: it was created by Claude, but I reviewed it thoroughly and made the important decisions together with Claude; in particular the one that COLUMNS takes precedence over `tput cols`. I think this makes sense because git does the same, and because it gives hosts a way to override the terminal width should they have a reason to do that (lazygit doesn't).

stefanhaller:
diff-so-fancy asks `tput cols` how wide to render. If $TERM is unset or names a terminal that terminfo does not know, tput fails; diff-so-fancy then prints a warning into its own output and renders at 80 columns. A caller that runs diff-so-fancy with no terminal on its output, such as an editor or a TUI showing the result in one of its own views, has no way to state a width.

Read $COLUMNS first, and shell out to tput only if $COLUMNS is unset or does not hold a positive number. Git resolves its own width the same way in term_columns(), and sets $COLUMNS for the pager it spawns, because by then its stdout is a pipe and the ioctl no longer reaches a terminal. diff-so-fancy.rulerWidth still wins over $COLUMNS.

An ncurses tput reads $COLUMNS itself, so on those systems the width was already right whenever tput ran at all. Reading $COLUMNS here makes it work with no $TERM as well, and saves the shell-out.

I didn't regenerate docs/diff-so-fancy.1; my local pod2man is older than the one the committed page came from, and would rewrite the whole file. It looks like the maintainer regenerates it separately anyway.

Fixes #555.


scottchiefbaker:
Seems sensible. 

Well written and designed PR. Clearly this isn't your first rodeo :)

